PR #18, opened 2020-05-26
Updated bulma to 0.8.2 and used is-light class for enabled tag

monika-Dianomic:
Before applying `is-light` class-
Screenshot 2020-05-26 at 2 52 42 PM

After applying `is-light` class-
Screenshot 2020-05-26 at 2 50 01 PM


MarkRiddoch:
Can we not keep the colour the same? I really we didn't change this as we would need to update the screenshots.

praveen-garg:
@monika-Dianomic let's revert `is-light` class added for South, North & Notifications page. Though GUI should have flexibility to do UI / UX improvements etc. but as we have screenshots in documentation, let's take that when there will be update in docs. and merge this into develop

PR #29, opened 2020-08-24
[FOGL-4420]: Removed repeat field for manual and startup schedule

monika-Dianomic:


src/app/components/core/scheduler/update-schedule/update-schedule.component.ts
```diff
@@ -168,14 +168,21 @@ export class UpdateScheduleComponent implements OnInit, OnChanges {
     }
   }
 
+  getRepeatTime() {
+    if (this.selectedScheduleTypeIndex === 1 || this.selectedScheduleTypeIndex === 4) {
```
praveen-garg, line 172: better to keep `**!** (selectedScheduleTypeIndex == 2 || selectedScheduleTypeIndex == 3) to keep same as in html template.

praveen-garg (changes requested):
There are few other issues too:
1) We should not see Repeat (Interval) for TIMED schedule. but Day / Time only
2) In list page, for TIMED don't show Interval Value but add a Time column as well and show Time value

src/app/components/core/scheduler/list-schedules/list-schedules.component.html
```diff
@@ -25,7 +26,13 @@
             </td>
             <td><small>{{schedule.type}}</small></td>
             <td>
-              <small *ngIf="schedule.repeat.trim() !== '00:00:00'"> {{schedule.repeat.trim()}}</small>
+              <small *ngIf="schedule.repeat.trim() !== '00:00:00' && schedule.type !== 'TIMED'"> {{schedule.repeat.trim()}}</small>
```
praveen-garg, line 29: should not this condition be,

`schedule.type !== 'INTERVAL'" && schedule.repeat.trim() !== '00:00:00' `?

praveen-garg (approved):

mshariq-nerd (approved):

PR #43, opened 2020-12-09
[FOGL-4721.patch2]: Sort by name on North page

monika-Dianomic:


src/app/components/core/north/north.component.ts
```diff
@@ -79,7 +79,7 @@ export class NorthComponent implements OnInit, OnDestroy {
         (data) => {
           this.tasks = data;
           this.tasks = sortBy(this.tasks, function (obj) {
-            return !obj.enabled + obj.processName.toLowerCase();
+            return !obj.enabled + obj.name.toLowerCase() + obj.processName.toLowerCase();
```
praveen-garg, line 82: Works as expected i.e. sorted by name and enabled first.

Why we had + obj.processName.toLowerCase(); earlier, were not all were north_c? 

and what impact this `obj.name.toLowerCase() + obj.processName.toLowerCase(); `
obj.processName.toLowerCase(); will make no difference b/w north_c and north_C.

praveen-garg (approved):

PR #46, opened 2020-12-14
[FOGL-4420.patch2]: DOM and css changes on Schedule and Configuration page

monika-Dianomic:
- [x] DOM fix on schedule page

- [x] css changes for the text "No sub-categories" on Configuration page

### DIFF
--- a/src/app/components/core/configuration-manager/configuration-manager.component.html
+++ b/src/app/components/core/configuration-manager/configuration-manager.component.html
@@ -25,12 +25,12 @@
             </div>
           </div>
         </div>
-        <div *ngIf="!isChild" class="panel-block">
-          <button class="button is-rounded is-fullwidth is-small" disabled>
-            no sub-category
-          </button>
+        <div *ngIf="!isChild" class="panel-block no-cat">
+          <div class="has-text-centered">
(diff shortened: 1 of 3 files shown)


src/app/components/core/configuration-manager/configuration-manager.component.html
```diff
@@ -25,12 +25,12 @@
             </div>
           </div>
         </div>
-        <div *ngIf="!isChild" class="panel-block">
-          <button class="button is-rounded is-fullwidth is-small" disabled>
-            no sub-category
-          </button>
+        <div *ngIf="!isChild" class="panel-block no-cat">
+          <div class="has-text-centered">
```
praveen-garg, line 29: `has-text-centered` class is required? we have `justify-content: center;`  on parent div. may be we can remove this div and keep small tag only.

praveen-garg (dismissed):

src/app/components/core/configuration-manager/configuration-manager.component.html
monika-Dianomic: fixed!

praveen-garg (approved):

PR #52, opened 2021-01-07
[FOGL-3921_gui]: Show connection info on settings page and show version on hover fledge logo

monika-Dianomic:
image

### DIFF
--- a/src/app/components/core/settings/settings.component.html
+++ b/src/app/components/core/settings/settings.component.html
@@ -47,8 +47,10 @@
               <!-- <button id="discover-fledge-btn" class="button is-text is-small" (click)="openServiceDiscoveryModal()">Discover Fledge Instances</button> -->
             </div>
           </div>
-          <div class="column is-2"></div>
-          <div *ngIf="version" class="column info no-rec"><small><b>Connected to</b> {{scheme}}://{{host}}:{{servicePort}}/fledge <b>version {{version}}</b>.</small></div>
+          <ng-container *ngIf="isServiceUp">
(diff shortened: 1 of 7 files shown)


src/app/components/core/settings/settings.component.html
```diff
@@ -47,6 +47,10 @@
               <!-- <button id="discover-fledge-btn" class="button is-text is-small" (click)="openServiceDiscoveryModal()">Discover Fledge Instances</button> -->
             </div>
           </div>
+          <ng-container *ngIf="version">
+            <div class="column is-2"></div>
```
praveen-garg, line 51: required?

praveen-garg (approved):

src/app/components/core/settings/settings.component.ts
```diff
@@ -21,6 +21,8 @@ export class SettingsComponent implements OnInit {
   refreshInterval: string;
   serviceUrl = '';
   selectedTheme: string;
+  version;
+  scheme;
```
mshariq-nerd, line 25: There is no need of another variable `scheme` as we already have global variable `protocol` for scheme.

mshariq-nerd (dismissed):

src/app/components/core/settings/settings.component.ts
monika-Dianomic: We have `protocol` variable in the way that it's value changed with the change in the protocol dropdrown, even without clicking save. So we can't use `protocol` variable here. cc/ @praveen-garg

mshariq-nerd (approved):

PR #58, opened 2021-02-09
[FOGL-2534.patch]: Fixes for the case when service down and ?id=1 removed from URL

monika-Dianomic:
- [x] Fixed console error on notifications menu option
- [x] Fixed empty red alert on other pages
- [x] Clear user session and redirect to login page after clicking logout button (when service is down)


src/app/services/http.request.interceptor.ts
```diff
@@ -27,10 +29,14 @@ export class HttpsRequestInterceptor implements HttpInterceptor {
         return event;
       }), catchError((err: any) => {
         if (err instanceof HttpErrorResponse) {
-          if (err.status === 401 && !location.href.includes('/setting?id=1')) {
+          if (err.status === 0) {
+            this.alertService.error('Failed to connect');
+          } else if (err.status === 401 && !location.href.includes('/setting?id=1')) {
             sessionStorage.clear();
             this.router.navigate(['/login']);
```
praveen-garg, line 36: this.servicesRecord = []; and replaceUrl: true  not needed here? as we have in clearUserSession func?

src/app/components/core/settings/settings.component.html
```diff
@@ -33,8 +33,11 @@
           </div>
           <div class="column">
             <label class="label is-small is-fullwidth">Port</label>
-            <input id='service_port' class="input" type="number" appNumberOnly min="0" maxlength="4" placeholder="8081"
-              [ngModel]="servicePort">
+            <input name="service_port" id='service_port' class="input" type="number" [minValue]="0" [maxValue]="65535" placeholder="8081"
+              [ngModel]="servicePort" #service_port="ngModel">
```
praveen-garg, line 37: what is `#service_port="ngModel"` next to `[ngModel]="servicePort"`

src/app/services/http.request.interceptor.ts
monika-Dianomic: `servicesRecord` doesn't exist on this page but `replaceUrl: true` can be used here.

src/app/components/core/settings/settings.component.html
monika-Dianomic: To get the value of `port` in local template variable, I have used it. We require this variable to show error message.

praveen-garg (approved):

mshariq-nerd (approved):

PR #72, opened 2021-03-10
FOGL-5081: Update application to Angular v11

mshariq-nerd:


package.json
```diff
@@ -8,7 +8,7 @@
     "prestart": "bash git-version",
     "start": "ng serve",
     "prebuild": "bash git-version",
-    "build": "ng build --prod --build-optimizer --vendor-chunk=true",
+    "build": "ng build --configuration production --build-optimizer --vendor-chunk=true",
```
praveen-garg, line 11: --build-optimizer --vendor-chunk=true are need here? or these shall go with aot etc, in config file for production.

praveen-garg (commented):
We better a requirements script as well for pre-requisite check and install; and in deploy call the ./build script

praveen-garg (commented):
docs/developers-guide.md needs an update along with supported browser versions.

package.json
mshariq-nerd: We can remove  --vendor-chunk as that is for dev purpose only as per angular doc 
> --vendor-chunk | Generate a separate bundle containing only vendor libraries. This option should only used for development.

mshariq-nerd: Read more https://angular.io/cli/build

PR #77, opened 2021-04-07
FOGL-5241: Documentation help icon is missing on right hand side of dropdown for notification delivery, rule & filters plugins

mshariq-nerd:


src/app/components/core/settings/settings.component.ts
```diff
@@ -23,7 +23,7 @@ export class SettingsComponent implements OnInit {
   selectedTheme: string;
   isServiceUp = false;
   version;
-  scheme;
+  scheme// default protocol
```
praveen-garg, line 26: fix lint

praveen-garg (approved):

monika-Dianomic (approved):

PR #79, opened 2021-04-08
FOGL-5260: Added filter in FogLAMP GUI to see foglamp-service-management logs only

mshariq-nerd:


src/app/components/core/system-log/system-log.component.ts
```diff
@@ -66,24 +66,33 @@ export class SystemLogComponent implements OnInit, OnDestroy {
...
+            if ('north_c' === sch.processName) {
               north_c.push(sch);
             }
-            if ('north_C'.includes(sch.processName)) {
+            if ('north_C' === sch.processName) {
               north_C.push(sch);
             }
-            if ('north'.includes(sch.processName)) {
+            if ('north' === sch.processName) {
               north.push(sch);
```
praveen-garg, line 86: We should have const north = []; only and push all 3 type of process name matching sch names to it.
And this.scheduleData = south_c.concat(north, notification_c, management, dispatcher_c);

src/app/components/core/system-log/system-log.component.ts
```diff
@@ -66,24 +66,33 @@ export class SystemLogComponent implements OnInit, OnDestroy {
...
             }
+            if ('management' === sch.processName) {
+              management.push(sch);
```
praveen-garg, line 89: ensure this shows captured logs of child process of this task manage / poll schedule

praveen-garg (approved):

src/app/components/core/system-log/system-log.component.ts
praveen-garg: Under Logs Task section. each timed / interval based task should also appear for poll schedule.

monika-Dianomic (approved):

PR #82, opened 2021-05-06
FOGL-5198: Fixed nested readings data parsing issue in downloaded CSV

mshariq-nerd:


src/app/services/generate-csv.service.ts
```diff
@@ -72,7 +73,7 @@ export class GenerateCsvService {
         return uniqueKeys.map(k => {
           const cell = k === 'timestamp' ? asset.timestamp :
             (asset.reading[k] === null || asset.reading[k] === undefined ? '' :
-              (asset.reading[k] instanceof Array ? '"' + asset.reading[k].join() + '"' : asset.reading[k]));
+              (asset.reading[k] instanceof Array ? '"' + asset.reading[k].join() + '"' : asset.reading[k] instanceof Object ? JSON.stringify(asset.reading[k]) : asset.reading[k]));
```
praveen-garg, line 76: csv download is okay. We should also see such data as is in Tabular tab (as textual data) when can not be rendered as a graph

praveen-garg: csv download is okay. We should also see such data as-is in Tabular tab (as textual data) when can not be rendered as a graph

https://github.com/fledge-iot/fledge-gui/pull/87

praveen-garg (approved):

monika-Dianomic (approved):

PR #89, opened 2021-05-21
FOGL-5402 simplify delete option in cert store. let API handled the configured certs and skipped login part

praveen-garg:


praveen-garg:
see https://github.com/fledge-iot/fledge/pull/392/files normal user should be able to import and overwrite other certs successfully, which are not (default shipped and configured for CA/ TLS). For skipped login, let API handle the response instead of hard coding.

monika-Dianomic (approved):

PR #92, opened 2021-05-26
[FOGL-5483_gui]: Endpoints changes in GUI w.r.t. FOGL-5483

monika-Dianomic:
Test with https://github.com/fledge-iot/fledge/pull/398

### DIFF
--- a/src/app/services/user.service.ts
+++ b/src/app/services/user.service.ts
@@ -68,12 +68,12 @@ export class UserService {
   /**
    * enable/disable user
    *
-   * PUT  | /fledge/admin/{id}/enabled
+   * PUT  | /fledge/admin/{id}/enable
    *
    * @param payload = > {"enabled": boolean}
    */
    enableUser(id, payload) {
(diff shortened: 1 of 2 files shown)


src/app/services/user.service.ts
```diff
@@ -113,7 +113,16 @@ export class UserService {
...
     };
-    return this.http.put(this.USER_URL + '/' + data.userId, payload).pipe(
+    return this.http.put(this.ADMIN_URL + '/' + data.userId, payload).pipe(
+      map(response => response),
+      catchError(error => throwError(error)));
+  }
+
+  updateMe(data) {
+    const payload: any = {
+      real_name: data.real_name
```
praveen-garg, line 123: test 
* a normal user cannot update role, access method, description of herself/ others.
* disabled user cannot log in
* deleted user cannot log in

and super admin (user id 1) is always there, cannot be deleted, role cannot be changed and only herself can update profile attributes (even not other admin)

monika-Dianomic: Tested all of the above, it seems Ok.

praveen-garg (approved):

PR #117, opened 2021-09-21
FOGL:5900: Filtering on logs difficult with large number of services

mshariq-nerd:
On System Logs page we have auto refresh functionality that refresh the data loaded in the page based on pingInterval. That causes issues with search functionality. So added option to start/stop auto refresh on system logs page

### DIFF
--- a/src/app/components/core/system-log/system-log.component.html
+++ b/src/app/components/core/system-log/system-log.component.html
@@ -72,7 +72,17 @@
           </div>
           <div class="column">
             <label class="label is-small">Search</label>
-            <input type="search" class="input is-fullwidth is-small" placeholder="search text on current page" [(ngModel)]="searchTerm" />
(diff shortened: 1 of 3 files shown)


src/app/components/core/system-log/system-log.component.html
```diff
@@ -90,8 +100,9 @@
     </ng-container>
     <ng-template #no_data_div>
       <div class="card-content">
-        <div class="has-text-centered"><small class="text-secondary has-text-grey-light">No Matching Log Entry Found</small></div>
+        <div class="has-text-centered"><small class="text-secondary has-text-grey-light">No Matching Log Entry
+            Found</small></div>
```
praveen-garg, line 104: No matching log entries found.

praveen-garg (dismissed):
The granularity on log pages will allow user to override the auto refresh behavior. User can disable and select/copy text OR continue tail with applied filter/search.

monika-Dianomic (approved):

PR #122, opened 2021-10-04
FOGL:5896: Error message is not showing on backup download

mshariq-nerd:


src/app/components/core/backup-restore/backup-restore.component.ts
```diff
@@ -191,16 +191,24 @@ export class BackupRestoreComponent implements OnInit, OnDestroy {
...
+      const date = this.dateFormatter.transform(backup.date, 'YYYY_MM_DD_HH_mm_ss');
+      a.download = 'fledge_backup_' + date + '.tar.gz';
+      document.body.appendChild(a);
+      a.click();
+      document.body.removeChild(a);
+    } catch (error) {
+      if (error.status === 0) {
+        console.log('service down ', error);
+      } else {
+        this.alertService.error(error.statusText);
```
praveen-garg, line 209: gradually the API is being improved to return JSON response body for error 4xx/5xx cases, not only the  status text (reason) in header. We should handle this gracefully and give priority to response error message and fallback to statusText.

praveen-garg (approved):

PR #131, opened 2021-11-10
FOGL-6069: Added doc link for notification type

mshariq-nerd:
Screen Shot 2021-11-10 at 11 41 39 AM

### DIFF
--- a/src/app/components/core/notifications/add-notification-wizard/add-notification-wizard.component.html
+++ b/src/app/components/core/notifications/add-notification-wizard/add-notification-wizard.component.html
@@ -108,8 +108,7 @@
               <div class="field">
                 <div class="control">
                   <ng-container *ngIf="notificationRulePlugins.length">
-                    <a class="button is-text text-btn"
-                      (click)="openPluginModal(true, 'Rule')">available plugins</a>
+                    <a class="button is-text text-btn" (click)="openPluginModal(true, 'Rule')">available plugins</a>
                   </ng-container>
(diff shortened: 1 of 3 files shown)


src/app/components/core/notifications/add-notification-wizard/add-notification-wizard.component.html
```diff
@@ -210,6 +208,11 @@
                   </div>
                 </div>
               </div>
+              <div
+                class="icon is-small is-tooltip-right tooltip is-pulled-right is-hovered help-icon notification-type-help"
+                data-tooltip="Help" (click)="goToNotificationTypeLink()">
+                <i class="far fa-question-circle"></i>
+              </div>
```
praveen-garg, line 215: For Retrigger Time description "A dead period (in seconds), during which notifications will not be sent regardless of the notification type."

see more around here. We may want to add same info in the doc page linked here.

mshariq-nerd: Right now for Retrigger time description we are showing `Retrigger time in seconds for sending a new notification` on tooltip. Do you mean to change that into `A dead period (in seconds), during which notifications will not be sent regardless of the notification type.`?

praveen-garg: Yes

praveen-garg (approved):

PR #192, opened 2022-07-25
FOGL-6661: Add developer feature to look at plugin data

mshariq-nerd:


MarkRiddoch (approved):
Looks good to me. Let's get this merged and open a new PR for when the API is ready that will allow the full list of plugins that persist data to be returned.

PR #194, opened 2022-08-01
FOGL-6613: Fledge GUI: Add support for ACL configuration item type

mshariq-nerd:
Test against https://github.com/fledge-iot/fledge/pull/738

### DIFF
--- a/src/app/components/core/configuration-manager/view-config-item/view-config-item.component.html
+++ b/src/app/components/core/configuration-manager/view-config-item/view-config-item.component.html
@@ -132,6 +132,15 @@
                       </div>
                     </div>
 
+                    <div *ngSwitchCase="'ACL'" class="select is-small">
+                      <select name='{{configItem.value.key}}' [ngModel]="setConfigValue(configItem.value)"
+                        [disabled]="configItem?.value?.hasOwnProperty('editable') && !configItem?.value?.editable"
+                        (change)="checkValidityOnChange(configItem.value, $event.target.value)">
+                        <option *ngFor="let configOption of controlAcls" [value]='configOption.name'>
+                          {{configOption?.name ? configOption?.name : 'None'}}</option>
+                      </select>
+                    </div>
+
                     <div *ngSwitchCase="'ENUMERATION'" class="select is-small">
                       <select name='{{configItem.value.key}}' [ngModel]="setConfigValue(configItem.value)"
                         [disabled]="configItem?.value?.hasOwnProperty('editable') && !configItem?.value?.editable"
--- a/src/app/components/core/configuration-manager/view-config-item/view-config-item.component.html
+++ b/src/app/components/core/configuration-manager/view-config-item/view-config-item.component.html
@@ -136,8 +136,8 @@
                       <select name='{{configItem.value.key}}' [ngModel]="setConfigValue(configItem.value)"
                         [disabled]="configItem?.value?.hasOwnProperty('editable') && !configItem?.value?.editable"
                         (change)="checkValidityOnChange(configItem.value, $event.target.value)">
-                        <option *ngFor="let configOption of controlAcls" [value]='configOption.name'>
-                          {{configOption?.name ? configOption?.name : 'None'}}</option>
+                        <option *ngFor="let acl of controlACLs" [value]='acl.name'>
+                          {{acl?.name ? acl?.name : 'None'}}</option>
                       </select>
                     </div>
 
--- a/src/app/components/core/configuration-manager/view-config-item/view-config-item.component.html
+++ b/src/app/components/core/configuration-manager/view-config-item/view-config-item.component.html
@@ -136,7 +136,7 @@
                       <select name='{{configItem.value.key}}' [ngModel]="setConfigValue(configItem.value)"
                         [disabled]="configItem?.value?.hasOwnProperty('editable') && !configItem?.value?.editable"
                         (change)="checkValidityOnChange(configItem.value, $event.target.value)">
-                        <option *ngFor="let acl of controlACLs" [value]='acl.name'>
+                        <option *ngFor="let acl of acls" [value]='acl.name'>
                           {{acl?.name ? acl?.name : 'None'}}</option>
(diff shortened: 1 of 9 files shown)


src/app/components/core/configuration-manager/view-config-item/view-config-item.component.html
```diff
@@ -132,6 +132,15 @@
                       </div>
                     </div>
 
+                    <div *ngSwitchCase="'ACL'" class="select is-small">
+                      <select name='{{configItem.value.key}}' [ngModel]="setConfigValue(configItem.value)"
+                        [disabled]="configItem?.value?.hasOwnProperty('editable') && !configItem?.value?.editable"
+                        (change)="checkValidityOnChange(configItem.value, $event.target.value)">
+                        <option *ngFor="let configOption of controlAcls" [value]='configOption.name'>
```
praveen-garg, line 139: let a | acl of acls

src/app/components/core/configuration-manager/view-config-item/view-config-item.component.ts
```diff
@@ -34,6 +35,7 @@ export class ViewConfigItemComponent implements OnInit, OnChanges, OnDestroy {
   public isValidForm = true;
   public isWizardCall = false;
   public filesToUpload = [];
+  public controlAcls = [];
```
praveen-garg, line 38: acls or ACLs

src/app/components/core/configuration-manager/view-config-item/view-config-item.component.ts
```diff
@@ -125,6 +133,21 @@ export class ViewConfigItemComponent implements OnInit, OnChanges, OnDestroy {
     }
   }
 
+  getAllACLs() {
+    this.controlService.fetchAllACL()
```
praveen-garg, line 137: We should probably move to ACL service

src/app/components/core/configuration-manager/view-config-item/view-config-item.component.ts
```diff
@@ -125,6 +133,21 @@ export class ViewConfigItemComponent implements OnInit, OnChanges, OnDestroy {
...
+      .subscribe((data: any) => {
+        this.controlAcls = orderBy(data.acls, 'name');
+        this.controlAcls.unshift({ name: '' }) // for empty acl value
```
praveen-garg, line 140: or empty ACL value, mapped to None text

src/app/components/core/configuration-manager/view-config-item/view-config-item.component.ts
```diff
@@ -108,6 +114,8 @@ export class ViewConfigItemComponent implements OnInit, OnChanges, OnDestroy {
 
       // check if editable config item found, based on readonly property
       for (const el of this.categoryConfiguration) {
+
+
```
praveen-garg, line 118: do we need these newlines?

pintomax:
@mshariq-nerd @praveen-garg As per specs in "Control Dispatcher Service" document

we need to call

ATTACH

PUT /fledge/service/$serviceName/ACL -d '{"acl_name" : "ACL1"}'

DETACH

DELETE 127.0.0.1:8081/fledge/service/$serviceName/ACL

The GUI calls instead, ACL item update for /fledge/$serviceNameSecurity
example
PUT http://192.168.56.10:8081/fledge/category/SIN2Security -d '{"ACL":"ACL1"}'


c/c @MarkRiddoch

deepanshu-yadav (approved):

mshariq-nerd:
@MarkRiddoch Current implementation is as per the JIRA requirement. We are getting ACL as a config item in the security category of the service. In GUI I have added a new configuration type ACL, that is rendered on the configuration page dynamically with other configuration item if any. When we change the config value, updating that using the API endpoint PUT fledge/category/<service>Security, that is true for any configuration item to update its value. 

Now, If we want to use API of attach/detach ACL with service we have to take it out from the configuration category and put it as a separate item of the service because in the current GUI design calling different api on the category configuration item update, we need to identify the ACL item in the config and separate it out then update it and resync the updated config.

Please advise.

MarkRiddoch:
> Now, If we want to use API of attach/detach ACL with service we have to take it out from the configuration category and put it as a separate item of the service because in the current GUI design calling different api on the category configuration item update, we need to identify the ACL item in the config and separate it out then update it and resync the updated config.
> 
> Please advise.

You don't need to take it out. You can very easily identify the ACL's, that is why they are a different type. We will use the ACL mechanism in other places than a configuration category, we must make sure we deal with ACL's in a consistent way, hence the attach and detach calls that need to be made in addition to the update of the configuration category. If you are telling me the GUI code is not up to the job of making two API calls then we will have to move the intelligence into the API rather than have the GUI do it.

mshariq-nerd:
@MarkRiddoch   I think Deepanshu doing the stuff that when config item is changed then notify to service? And we are getting ACL type in Security child category for this purpose?
if we need to make two calls based on ACL, I can chain that. let me know. Security category config item update call will be there anyway?

This is how we are doing it in GUI right now.
https://user-images.githubusercontent.com/16384273/182409925-072bbcc8-bcf4-42c8-9093-3dad34765cde.mov

pintomax:
@mshariq-nerd when we select an ACL in the GUI, say ACL1 then
the GUI should do:

ATTACH

PUT /fledge/service/$serviceName/ACL -d '{"acl_name" : "ACL1"}'


Wen we select "none"

DETACH

DELETE 127.0.0.1:8081/fledge/service/$serviceName/ACL

Those are the only two APIs for adding/removing an ACL to a service

If an ACL is already set, we can only remove it.

Then a new one should be attached

praveen-garg (approved):

praveen-garg:
This looks good to move forward.

mshariq-nerd:
As this base work is required and we are able to display and save ACL type config item; I’m merging this.
@MarkRiddoch, please add a Jira for any change/extension required to this, with a clarification note to my earlier comment for my overall understanding.

PR #156, opened 2022-02-14
semver version info on settings page and removal of next from package naming

mshariq-nerd:

PR #213, opened 2022-10-11
fixed broken link for control script help doc

mshariq-nerd:


praveen-garg:
patch for https://github.com/fledge-iot/fledge-gui/pull/156

src/app/services/doc.service.ts
```diff
@@ -8,33 +8,30 @@ import * as data from '../../git-version.json';
 export class DocService {
   public gitDistance = data['default'].distance;
   public appVersion = data['default'].tag;
-
-  constructor() { }
+  public version = 'develop';
+  constructor() {
+    this.version = this?.gitDistance > 0 ? 'develop' : `${this?.appVersion}`;
```
praveen-garg, line 13: do we really need string interpolation here? and `this?`

praveen-garg (approved):

src/app/services/doc.service.ts
praveen-garg: I suggest to make this

```
  private version = "develop"; //readthedocs version to use

  constructor() {
    if (this.gitDistance == 0) {
      this.version = this.appVersion;
    }
  }
```
so that in case a custom branch doc tests needed (say 2.0.1RC), GUI can be tested with single change

praveen-garg (approved):

PR #223, opened 2022-11-14
[FOGL-7085]: Fixed issue of assets and readings chart for 1 week

MonikaSharma06:


src/app/components/core/asset-readings/readings-graph/readings-graph.component.ts
```diff
@@ -621,7 +621,7 @@ export class ReadingsGraphComponent implements OnDestroy {
           distribution: 'linear',
           type: 'time',
           time: {
-            unit: 'second',
+            unit: this.optedTime !== 604800 ? 'second' : '',
             tooltipFormat: 'YYYY-MM-DD HH:mm:ss.SSS',
             displayFormats: {
               unit: 'second',
```
praveen-garg, line 627: What is this unit for? And does this matter for a week data rendering vs lesser?

monika-Dianomic: If `unit` is defined, it force the unit to be a certain type. For example- We use `unit: 'year'` to create a chart with a time scale that always displayed units per year.

Here, I had to remove unit option to fix the issue. By applying condition, I see issue is still there.

PR #231, opened 2022-12-05
FOGL-7108: role based access for view & data_view role

mshariq-nerd:
Test with https://github.com/fledge-iot/fledge/pull/899

### DIFF
--- a/src/app/components/core/backup-restore/backup-restore.component.html
+++ b/src/app/components/core/backup-restore/backup-restore.component.html
@@ -20,15 +20,17 @@
               </div>
               <div class="dropdown-menu" id="dropdown-menu3" role="menu">
                 <div class="dropdown-content">
-                  <a class="dropdown-item" id="create-backup" (click)="requestBackup()">
-                    <i class="far fa-plus-square fa-sm fa-fw" aria-hidden="true"></i>
-                    <span> Create New</span>
-                  </a>
-                  <a class="dropdown-item" id="upload-backup" (click)="uploadBackup()">
-                    <i class="fas fa-paperclip fa-sm fa-fw" aria-hidden="true"></i>
-                    <span> Upload a Backup</span>
-                  </a>
-                  <hr class="dropdown-divider">
+                  <ng-container *ngIf="rolesService.hasEditorRole()">
+                    <a class="dropdown-item" id="create-backup" (click)="requestBackup()">
+                      <i class="far fa-plus-square fa-sm fa-fw" aria-hidden="true"></i>
+                      <span> Create New</span>
+                    </a>
+                    <a class="dropdown-item" id="upload-backup" (click)="uploadBackup()">
+                      <i class="fas fa-paperclip fa-sm fa-fw" aria-hidden="true"></i>
+                      <span> Upload a Backup</span>
+                    </a>
+                    <hr class="dropdown-divider">
+                  </ng-container>
                   <a class="dropdown-item" (click)="goToLink()">
                     <i class="far fa-question-circle fa-sm fa-fw"></i>
                     <span> Help</span>
@@ -81,12 +83,13 @@
                   <small>{{ backup.status }}</small>
                 </td>
                 <td class="align-content">
-                  <a *ngIf="backup.status != 'RUNNING'" class="button is-text text-btn"
+                  <a *ngIf="backup.status != 'RUNNING' && rolesService.hasEditorRole()" class="button is-text text-btn"
                     (click)="openModal(backup.id, backup.date, 'Are you sure you want to delete the backup', 'deleteBackup')">delete
                   </a>
                 </td>
                 <td class="align-content">
-                  <a *ngIf="backup.status == 'COMPLETED' || backup.status == 'RESTORED'" class="button is-text text-btn"
+                  <a *ngIf="(backup.status == 'COMPLETED' || backup.status == 'RESTORED') && rolesService.hasEditorRole()"
+                    class="button is-text text-btn"
                     (click)="openModal(backup.id, backup.date, 'Are you sure you want to restore the backup', 'restoreBackup')">restore
                   </a>
                 </td>
--- a/src/app/components/core/backup-restore/backup-restore.component.html
+++ b/src/app/components/core/backup-restore/backup-restore.component.html
@@ -20,7 +20,7 @@
               </div>
               <div class="dropdown-menu" id="dropdown-menu3" role="menu">
                 <div class="dropdown-content">
-                  <ng-container *ngIf="rolesService.hasEditorRole()">
+                  <ng-container *ngIf="rolesService.hasEditPermissions()">
                     <a class="dropdown-item" id="create-backup" (click)="requestBackup()">
                       <i class="far fa-plus-square fa-sm fa-fw" aria-hidden="true"></i>
                       <span> Create New</span>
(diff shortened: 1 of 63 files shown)


src/app/components/layout/side-menu/side-menu.component.html
```diff
@@ -9,89 +9,92 @@
...
+        Support
+      </a>
+    </li>
+    <li>
+      <a id="developer" routerLinkActive="is-active" routerLink='/developer'
+        *ngIf="developerFeaturesService.getDeveloperFeatureControl() &&  rolesService.hasEditorRole()">
+        Developer
+      </a>
+    </li>
+  </article>
```
praveen-garg, line 97: Why an _article_ tag?

src/app/directives/access-control.directive.ts
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  ngOnChanges(): void {
+    this.validateAccess();
+  }
+
+  private validateAccess(): void {
+    const roleId = Number(sessionStorage.getItem('roleId'));
+    if (roleId === 4) {
+      // remove action element from DOM for data_view user
```
praveen-garg, line 23: which action element ? Add What role name is for the ID 4?

src/app/guards/roles.gurad.ts
```diff
@@ -0,0 +1,16 @@
+import { Injectable } from '@angular/core';
+import { CanActivate, Router } from '@angular/router';
+import { RolesService } from '../services';
+
+@Injectable()
+export class RolesGuard implements CanActivate {
+  constructor(private router: Router, public rolesService: RolesService) { }
+  canActivate() {
+    const isEditor = this.rolesService.hasEditorRole();
```
praveen-garg, line 9: const canEdit = this.rolesService.hasEditPermissions();

src/app/services/roles.service.ts
```diff
@@ -0,0 +1,30 @@
...
+    return [appRoles.admin, appRoles.user, appRoles.anonymous].includes(roleId);
+  }
+
+  /**
+   * To check if user have data_view role
+   * @returns true|false based on user role
+   */
+  public hasDataViewRole(): boolean {
+    const roleId = Number(sessionStorage.getItem('roleId'));
+    console.log('role', roleId == 4);
```
praveen-garg, line 26: remove?

src/app/services/roles.service.ts
```diff
@@ -0,0 +1,30 @@
...
+})
+export class RolesService {
+
+  /**
+   * To check if user, who has the add/edit permission
+   * @returns true|false based on user role
+   */
+  public hasEditorRole(): boolean {
+    const roleId = Number(sessionStorage.getItem('roleId'));
+
```
praveen-garg, line 16: remove extra line?

src/app/services/roles.service.ts
```diff
@@ -0,0 +1,30 @@
+import { Injectable } from '@angular/core';
+
+enum appRoles { anonymous = 0, admin = 1, user = 2, view = 3, data_view = 4 };
```
praveen-garg, line 3: we can have this object from API call?

praveen-garg: It would be better to show Role name in GUI as:

1. Administrator | Admin 
2. Editor | Edit 
3. Viewer | View
4. Data Viewer | Data View

src/app/components/core/backup-restore/backup-restore.component.html
```diff
@@ -81,12 +83,13 @@
...
                 <td class="align-content">
-                  <a *ngIf="backup.status != 'RUNNING'" class="button is-text text-btn"
+                  <a *ngIf="backup.status != 'RUNNING' && rolesService.hasEditorRole()" class="button is-text text-btn"
                     (click)="openModal(backup.id, backup.date, 'Are you sure you want to delete the backup', 'deleteBackup')">delete
                   </a>
                 </td>
                 <td class="align-content">
-                  <a *ngIf="backup.status == 'COMPLETED' || backup.status == 'RESTORED'" class="button is-text text-btn"
+                  <a *ngIf="(backup.status == 'COMPLETED' || backup.status == 'RESTORED') && rolesService.hasEditorRole()"
+                    class="button is-text text-btn"
```
praveen-garg, line 92: can have ngContainer?

src/app/services/roles.service.ts
```diff
@@ -0,0 +1,30 @@
...
+@Injectable({
+  providedIn: 'root'
+})
+export class RolesService {
+
+  /**
+   * To check if user, who has the add/edit permission
+   * @returns true|false based on user role
+   */
+  public hasEditorRole(): boolean {
```
praveen-garg, line 14: I suggest to rename this function as `hasEditPermissions()`

src/app/services/roles.service.ts
mshariq-nerd: >we can have this object from API call?
If we will do that, either we need to fetch all roles every time in this service and map, that to me is not a good idea as  we guards the routes based on role and in role guards we are accessing this service to know the role,
Else we need to hold all the roles in the session that we are fetching on login? 


* To change the naming on UI better to do it using new JIRA.

src/app/components/core/south/south-service-modal/south-service-modal.component.html
```diff
@@ -15,15 +15,15 @@
           <app-config-children #configChildComponent *ngIf="category" [category]="category"
             [plugin]="service?.plugin.name" [serviceStatus]="service['schedule_enabled']" [from]="'south'">
           </app-config-children>
-          <div *ngIf="service != undefined" class="columns">
+          <div *ngIf="service" class="columns">
```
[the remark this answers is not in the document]
mshariq-nerd: No, `*ngIf service`  check do the same.

src/app/components/layout/side-menu/side-menu.component.html
mshariq-nerd: For `data_view` role we need to hide the sidebar menu items those I am hiding based on a `requiredViewEditorRole` directive and to apply a directive a DOM element required.
we are hiding many elements in the side menu so instead of applying directive on every element, I added `<article requiredViewEditorRole>`. 

Also The <article> element does not render as anything special in a browser.

src/app/components/core/settings/settings.component.html
```diff
@@ -190,7 +190,7 @@
       </div>
       <hr>
       <div class="columns">
-        <div class="column is-6">
+        <div class="column is-6" *ngIf="rolesService.hasEditorRole()">
```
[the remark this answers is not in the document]
mshariq-nerd: JIRA says 
>For both types(view & data_view) only users
Remove developers options toggle from the settings page and make sure develop options are turned off

But the developer tab in sidebar menu only get enabled when developer option set true in settings page.

src/app/services/roles.service.ts
praveen-garg: OK, please file a Jira and add notes which dropdown need binding to these value as options.

praveen-garg: I wonder if needed the API call then which we requested for data view, if we have to hard-code these. Obviously anonymous = 0 needs to be augmented.

src/app/components/core/settings/settings.component.html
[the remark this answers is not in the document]
praveen-garg: OK, Please add a note to Jira that View User shall not be able to View Python Packages, As Developer options sidebar menu will be hidden.

Mohit04tomar (approved):

praveen-garg (approved):

PR #232, opened 2022-12-16
FOGL-7193: collapse assets & readings by default on south page 

mshariq-nerd:


src/app/components/core/south/south.component.ts
```diff
@@ -101,7 +101,7 @@ export class SouthComponent implements OnInit, OnDestroy {
           this.southboundServices = orderBy(enabledServices, 'name').concat(orderBy(disabledServices, 'name'));
           // add expanded key in service to show/hide the assets in the service row
           this.southboundServices.map((svc: any) => {
-            svc.expanded = true;
+            svc.expanded = false;
```
praveen-garg, line 104: Are we storing this state in session storage?

mshariq-nerd: No we are not. Issue explained in JIRA is when user go back and forth from south to other page, coming back on south page the assets always appear as expanded. Now those will be collapsed by default.

praveen-garg: OK

monika-Dianomic (approved):

PR #239, opened 2022-12-29
FOGL-7307: Fixed duplicate name issue for different categories in group

mshariq-nerd:


praveen-garg (approved):
This looks good!

Validity expression check on Advanced/Security Categories can be done via a new Jira; I don't think we have any validity rules there.

PR #246, opened 2023-01-20
FOGL-7375: Enable/Disable failed to save for filter-python35

mshariq-nerd:
double quote empty string check added when creating file to upload with value


praveen-garg (changes requested):
No, This is not a GUI issue `""""` is not empty `""`

PR #249, opened 2023-02-02
[FOGL-7191]: Restrict entry of double quote in the names of services, tasks, filters, notification, scripts and ACLs

monika-Dianomic:


src/app/directives/restrict-double-quote.directive.ts
```diff
@@ -0,0 +1,24 @@
+import { Directive, HostListener } from '@angular/core';
+
+@Directive({
+  selector: '[appRestrictDoubleQuote]'
+})
+export class RestrictDoubleQuoteDirective {
+
+  constructor() { }
```
praveen-garg, line 8: @Input('chars') chars: string  = "34"; // a comma separated list of special char code values 

... and default is 34

src/app/directives/restrict-double-quote.directive.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  @HostListener('keypress') onkeypress(e) {
+    const event = e || window.event;
+    if (event) {
+      return this.isDoubleQuote(event);
```
praveen-garg, line 13: return this.isRestrictedChar(event);

src/app/directives/restrict-double-quote.directive.ts
```diff
@@ -0,0 +1,24 @@
...
+    }
+  }
+
+  isDoubleQuote(event) {
```
praveen-garg, line 17: isRestrictedChar

src/app/directives/restrict-double-quote.directive.ts
```diff
@@ -0,0 +1,24 @@
...
+    const charCode = (event.which) ? event.which : event.keyCode;
+    if (charCode === 34) {
+      return false;
+    }
+    return true;
```
praveen-garg, line 22: return charCode in cc

src/app/directives/restrict-special-char.directive.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  @HostListener('keypress') onkeypress(e) {
+    const event = e || window.event;
+    if (event) {
+      return this.isRestrictedChar(event);
+    }
+  }
+
+  isRestrictedChar(event) {
+    if (!this.restrictedChars) {
```
praveen-garg, line 19: ! or not array or array length empty

praveen-garg (approved):
I assume we cannot update the name hence no such need.

Also check Username?
We can add a Jira for other form fields and modules as we see for character like ` " ' \t \r \n etc.

src/app/components/core/control-dispatcher/add-control-acl/add-control-acl.component.html
```diff
@@ -29,7 +29,7 @@
             <div class="field-body">
               <div class="field">
                 <div class="control">
-                  <input name="name" class="input" type="text" placeholder="Name" required [(ngModel)]="name"
+                  <input name="name" class="input" type="text" placeholder="Name" required [(ngModel)]="name" appRestrictSpecialChar
```
praveen-garg, line 32: appRestrictSpecialChars

src/app/components/core/south/add-service-wizard/add-service-wizard.component.html
```diff
@@ -83,7 +83,7 @@
             <div class="field-body">
               <div class="field">
                 <div class="control">
-                  <input class="input" id="name" type="text" placeholder="name" formControlName="name" required
+                  <input class="input" id="name" appRestrictSpecialChar [restrictedCharCode]="[34]" type="text" placeholder="name" formControlName="name" required
```
praveen-garg, line 86: No need to pass [restrictedCharCode]="[34]"

src/app/directives/restrict-special-char.directive.ts
```diff
@@ -0,0 +1,25 @@
+import { Directive, HostListener, Input } from '@angular/core';
+
+@Directive({
+  selector: '[appRestrictSpecialChar]'
+})
+export class RestrictSpecialCharDirective {
+  @Input() restrictedCharCode: any[];
```
praveen-garg, line 7: restrictedCharCodes

src/app/directives/restrict-special-char.directive.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  @HostListener('keypress') onkeypress(e) {
+    const event = e || window.event;
+    if (event) {
+      return this.isRestrictedChar(event);
+    }
+  }
+
+  isRestrictedChar(event) {
+    if (this.restrictedCharCode.length === 0) {
```
praveen-garg, line 19: this.restrictedCharCodes?.length

praveen-garg: what happens if we pass [restrictedCharCode]="" or [restrictedCharCode]="34, b"

monika-Dianomic: > what happens if we pass [restrictedCharCode]="" or [restrictedCharCode]="34, b"

`[restrictedCharCode]=""` will allow all characters but `[restrictedCharCode]="34, b"` will give error. If we use `[restrictedCharCode]="[34, b]"` instead of `[restrictedCharCode]="34, b"`, it will restrict entry of `"`.

src/app/directives/restrict-special-char.directive.ts
```diff
@@ -0,0 +1,25 @@
...
+  @HostListener('keypress') onkeypress(e) {
+    const event = e || window.event;
+    if (event) {
+      return this.isRestrictedChar(event);
+    }
+  }
+
+  isRestrictedChar(event) {
+    if (this.restrictedCharCodes?.length === 0) {
+      this.restrictedCharCodes = ['34'];
```
mshariq-nerd, line 20: Why `34` as string in the array? It is failing in this.restrictedCharCodes?.includes(charCode); because event.keyCode is a number. 


you can make it more simple. like
```
restrictedCharCodes = [34];  // 34 = double quote(") key code

  constructor() { }

  @HostListener('keypress') onkeypress(e) {
    const event = e || window.event;
    if (event) {
      return this.isRestrictedChar(event);
    }
  }

  isRestrictedChar(event) {
    const charCode = (event.which) ? event.which : event.keyCode;
    return !this.restrictedCharCodes?.includes(charCode);
  }
```

src/app/directives/restrict-special-char.directive.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  @HostListener('keypress') onkeypress(e) {
+    const event = e || window.event;
+    if (event) {
+      return this.isRestrictedChar(event);
+    }
+  }
+
+  isRestrictedChar(event) {
+    if (this.restrictedCharCodes?.length === 0) {
```
mshariq-nerd, line 19: Is this check really required?

src/app/directives/restrict-special-char.directive.ts
```diff
@@ -0,0 +1,25 @@
+import { Directive, HostListener, Input } from '@angular/core';
+
+@Directive({
+  selector: '[appRestrictSpecialChars]'
+})
+export class RestrictSpecialCharDirective {
+  @Input() restrictedCharCodes: any[];
```
mshariq-nerd, line 7: What is the purpose of @Input() ?

mshariq-nerd (changes requested):

src/app/directives/restrict-special-char.directive.ts
monika-Dianomic: > Why 34 as string in the array?

Right, it should be a number. Actually, that was mistakenly written as string in the previous commit. It was a number before. I'll fix it in the next commit.

> It is failing in this.restrictedCharCodes?.includes(charCode); because event.keyCode is a number.

No, It was failing because `restrictedCharCodes` has undefined value (it is not entering in the `if` code block). The reason behind this is that- for now we are not passing any value to `restrictedCharCodes` from any parent component.
I'll add a check for this.

After reading my above answers, it will be clear to you that I can't change the code as above (in your comment).

src/app/directives/restrict-special-char.directive.ts
monika-Dianomic: Yes, it is needed as `restrictedCharCodes` is an array and it's value can be passed as `[]`

src/app/directives/restrict-special-char.directive.ts
monika-Dianomic: We can pass any value to `restrictedCharCodes` from any component and its value will be an array of the character codes which we want to restrict to enter in any input.

mshariq-nerd: As you said `restrictedCharCodes` is always an array then you can declare restrictedCharCodes as `@Input() restrictedCharCodes = [];`. By doing this you can remove the undefined check in if block on line 19.

src/app/directives/restrict-special-char.directive.ts
```diff
@@ -0,0 +1,25 @@
...
+  @HostListener('keypress') onkeypress(e) {
+    const event = e || window.event;
+    if (event) {
+      return this.isRestrictedChar(event);
+    }
+  }
+
+  isRestrictedChar(event) {
+    if (!this.restrictedCharCodes || this.restrictedCharCodes?.length === 0) {
+      this.restrictedCharCodes = [34];
```
mshariq-nerd, line 20: Please add a comment on this line why it is hardcoded here.

mshariq-nerd (changes requested):

mshariq-nerd (approved):

praveen-garg (approved):

PR #260, opened 2023-02-23
FOGL-7498: show blank value for config items if user changed it to blank

monika-Dianomic:


src/app/services/configuration-control.service.ts
```diff
@@ -145,7 +145,7 @@ export class ConfigurationControlService {
     Object.keys(configuration).forEach(key => {
       const element = configuration[key];
       element.key = key;
-      element.value = element.value ? element.value : element.default;
+      element.value = (element.value || element.value === '') ? element.value : element.default;
```
praveen-garg, line 148: should we have (element.value || element?.value === '')

praveen-garg: or much safer 
element.value !== undefined  ? element.value : element.default;

gnandan (commented):
is element.key always string type? if not, is it possible to update element.value only for string types ?

monika-Dianomic:
@gnandan We always get value cast as string.

PR #265, opened 2023-02-27
FOGL-7509: greyed out menu items when fledge is down

mshariq-nerd:


monika-Dianomic (dismissed):

praveen-garg (dismissed):
We can probably add opacity to container and also the open modal if any.  
We should not close the modal though as it can be temporarily down and may have to attempt 3 times if want to close the modal.

src/app/components/layout/navbar/navbar.component.css
```diff
@@ -86,6 +86,13 @@
...
-  cursor: default;
-}
+.disabled {
+  opacity: .95;
+  background-color: whitesmoke;
+  pointer-events: none;
+ }
+
+ .help {
+  margin-top: 0px;
```
praveen-garg, line 96: Not for any with help class but for this ID only?!

mshariq-nerd: help class is used only in the span that is to show the Auto refresh not enable message in navbar.

hemantgarg923 (approved):

src/app/app.component.css
```diff
@@ -0,0 +1,13 @@
+.c1 {
```
monika-Dianomic, line 1: I think, class name should be meaningful.

monika-Dianomic (approved):

PR #267, opened 2023-03-01
FOGL-7489: Able to edit disabled codemirror field issue fixed

hemantgarg923:
CASE-1
when validity is not available for script field then code is setting disability according to the availability of file.
CASE-2
when validity is available then it is setting disability according to validity expression i. e. it is not checking the availability of file in case of validity expression.


mshariq-nerd:
This will be closed in favour of https://github.com/fledge-iot/fledge-gui/pull/285

mshariq-nerd:
Closed in favour of https://github.com/fledge-iot/fledge-gui/pull/285

PR #275, opened 2023-03-27
FOGL-7555: show graph for additional assets

MonikaSharma06:


mshariq-nerd (approved):

hemantgarg923 (approved):
Working fine in all cases other than 3D graph. Whenever an asset is added containing 3D graph (example - FFT filter) simultaneously  asset dropdown disappears and user is unable to add further assets.
But this can be fixed in a separate PR and current PR can be merged.

PR #299, opened 2023-05-31
fixed broken css property of 'button is-small' class

monika-Dianomic:


src/app/components/core/asset-readings/readings-graph/readings-graph.component.css
```diff
@@ -12,11 +12,7 @@
   padding: 15px !important;
 }
 
-::ng-deep .button.is-small {
-  font-size: 0.90rem !important;
-}
-
-::ng-deep .input.is-small {
+::ng-deep .button.small-btn {
```
mshariq-nerd, line 15: why ng-deep?

src/app/components/common/time-dropdown/time-dropdown.component.html
```diff
@@ -5,7 +5,7 @@
 
   <div id="unit-dropdown" class="dropdown control column is-5 no-padding">
     <div class="dropdown-trigger">
-      <button class="button is-small has-background-white-ter unit-button" aria-haspopup="true"
+      <button class="button is-small small-btn has-background-white-ter unit-button" aria-haspopup="true"
```
mshariq-nerd, line 8: no supporting css for `.small-btn` in time-dropdown.component.css?

mshariq-nerd (changes requested):

src/app/components/core/asset-readings/readings-graph/readings-graph.component.css
monika-Dianomic: We are using `time-dropdown` component in readings-graph page as a child component. This css is used to apply css property on `small-btn` class of it's child component `time-dropdown`. So `ng-deep` is required here.

monika-Dianomic: I think, this did not break anything. It was breaking before (in develop) which I fixed here. As you can see, I have used custom class `small-btn` which is not being used anywhere. And for safest part, I have also used `:host` before `::ng-host` now.

> why we need to do it? Can't we put the child component css in the css class of the child?

As I told in the reply of another comment, I haven't applied css property in child component css because it may be child component of many other components where we may not need same property.

src/app/components/core/north/north-task-modal/north-task-modal.component.html
```diff
@@ -83,7 +83,7 @@
             <div class="column has-text-centered">
               <span class="label is-size-6 add-application" *ngIf="rolesService.hasEditPermissions(); else staticLabel">
                 <a (click)="openAddFilterModal(true)">Applications
-                  <span class="icon">
+                  <span class="icon" title="Add Filters">
```
praveen-garg, line 86: For consistency, This should be tooltip or title only?

src/app/components/core/north/north.component.html
```diff
@@ -3,7 +3,7 @@
     <header class="card-header">
       <div id="north-tasks" class="card-header-title">
         North Instances
-        <button *ngIf="!isAlive" (click)="getNorthTasks(true)" class="button is-small" id="refresh-check">
+        <button *ngIf="!isAlive" (click)="getNorthTasks(true)" class="button is-small" title="Reload" id="refresh-check">
```
praveen-garg, line 6: Add for all cards + Asset readings; note: for warning boxes of service added/installed this will be Recheck. Reload + Recheck can be title only when not in group

praveen-garg (approved):
Fwe minor comments

hemantgarg923 (approved):

mshariq-nerd (approved):

PR #302, opened 2023-06-06
FOGL-7833: Route guards updated according to authentication mechanism and user role

hemantgarg923:


src/app/guards/valid-user.guard.ts
```diff
@@ -2,14 +2,14 @@ import { Injectable } from '@angular/core';
...
   constructor(private router: Router) { }
   canActivate() {
-    if (sessionStorage.getItem('token') || sessionStorage.getItem('LOGIN_SKIPPED')) {
-      // logged in so return true
+    if (sessionStorage.getItem('token') || sessionStorage.getItem('LOGIN_SKIPPED') == 'true') {
+      // return true for (auth=optional) or (auth=mandatory and user role=any)
       return true;
     }
-    // not logged in so redirect to login page
+    // redirect to login for auth=mandatory and without any user token i.e. trying to access from url
```
praveen-garg, line 12: access pages URL

src/app/guards/valid-user.guard.ts
```diff
@@ -2,14 +2,14 @@ import { Injectable } from '@angular/core';
 import { CanActivate, Router } from '@angular/router';
 
 @Injectable()
-export class AuthCheckGuard implements CanActivate {
+export class ValidUserGuard implements CanActivate {
```
praveen-garg, line 5: AuthCheckGuard was better itself. I suggest to rename to AuthRequiredGuard

hemantgarg923: I think `AuthRequiredGuard` may not be that much intuitive here because the guard will return true for anonymous user also, and we already have `AuthGuard` for checking authentication.

praveen-garg: AuthRequiredGuard should return false if user can access anonymously or already authenticated and has token; 


If auth is mandatory (required) and no token, then return true.

mshariq-nerd (dismissed):

mshariq-nerd (approved):

monika-Dianomic (approved):

praveen-garg (approved):

PR #306, opened 2023-06-26
FOGL-7783: Dropdown created to see the graph from most recent reading for an asset

hemantgarg923:


MarkRiddoch (changes requested):
Having a dropdown with options of 'now' and 'last reading' does not seem like a good user interface, especially as we have the 'video buttons' for moving back and forth in time. Now may not always mean now, it depends on the interaction with those buttons that were added previously. A better user experience would be to extend those buttons by adding an extra one that move to to the time of the last reading. Think of it as a convenience function that moves you back along the time line to the point that readings were last received for the asset.

src/app/services/assets.service.ts
```diff
@@ -43,7 +43,7 @@ export class AssetsService {
       catchError(error => throwError(error)));
   }
 
-  public getMultipleAssetReadings(assetCode, limit: number = 0, offset: number = 0, time: number = 0, additionalAssets, previous: number = 0) {
+  public getMultipleAssetReadings(assetCode, limit: number = 0, offset: number = 0, time: number = 0, additionalAssets, previous: number|string = 0, mostrecent:boolean = false) {
```
praveen-garg, line 46: previous must remain a number; duration seconds

src/app/components/core/asset-readings/readings-graph/readings-graph.component.css
```diff
@@ -181,6 +181,14 @@ table.is-borderless td {
   float: right;
 }
 
+.btn-enabled {
+  color: #3850b7;
+}
+
+.btn-enabled {
```
praveen-garg, line 188: duplicate CSS code?

src/app/components/core/asset-readings/readings-graph/readings-graph.component.ts
```diff
@@ -59,7 +60,9 @@ export class ReadingsGraphComponent implements OnDestroy {
   public infoTextTimestamps = { start: "", end: "" };
   public graphStartTimestamp: string;
   public zoomConfig = { minZoomValue: 1, isZoomed: false };
+  public fromMostRecent: boolean = false;
   public isReadingsFetched = false;
+  public mostRecentReadingTimestamp;
```
MonikaSharma06, line 65: type?

src/app/components/core/asset-readings/readings-graph/readings-graph.component.ts
```diff
@@ -968,17 +974,28 @@ export class ReadingsGraphComponent implements OnDestroy {
...
-  showForwardReadingsGraphOrSummary() {
-    this.backwardReadingCounter--;
-    this.showReadingsGraphOrSummary();
+    if (!this.fromMostRecent) {
+      this.showReadingsGraph();
+      return;
+    }
+    let mostRecentReadingTimestampDate = new Date(this.mostRecentReadingTimestamp);
+    let prev = new Date(mostRecentReadingTimestampDate.valueOf() - this.backwardReadingCounter * this.optedTime * 1000);
+    let previous_ts = (moment(prev.valueOf()).format('YYYY-MM-DD HH:mm:ss.SSS')) + this.mostRecentReadingTimestamp.slice(-3); // send previous_ts upto 6 digits i.e. microsecond precision
```
MonikaSharma06, line 994: Please explain more about `previous_ts` format and its calculation.
While testing it, I'm getting it as `previous_ts = 2024-05-30%2001:52:39.934951`, here date it clear to me and microsecond is at last i.e. `934951` but what is 2001 ?

hemantgarg923: We are using `YYYY-MM-DD HH:mm:ss.SSS` timestamp format for previous_ts, here `%20` is for space and `01` is hours

mshariq-nerd (approved):

hemantgarg923:
> Having a dropdown with options of 'now' and 'last reading' does not seem like a good user interface, especially as we have the 'video buttons' for moving back and forth in time. Now may not always mean now, it depends on the interaction with those buttons that were added previously. A better user experience would be to extend those buttons by adding an extra one that move to to the time of the last reading. Think of it as a convenience function that moves you back along the time line to the point that readings were last received for the asset.

@MarkRiddoch Above requested changes are done as one more button is added to view the graph from most recent reading ingested in the system, please have a look.

MonikaSharma06 (approved):

MarkRiddoch (approved):
Looks good to me

PR #310, opened 2023-06-27
FOGL-7906: Do not display empty "Basic" tab

mshariq-nerd:


src/app/components/core/configuration-manager/configuration-group/configuration-group.component.html
```diff
@@ -38,33 +38,40 @@ <h6 class="is-6 is-small-font">
...
-      <app-show-configuration [hidden]="selectedGroup !== g.group" [groupConfiguration]="g.config" [group]="g.group" [from]="from"
-        [selectedGroup]="selectedGroup" [fullConfiguration]="category.config" (event)="getChangedConfiguration($event)"
-        (formStatusEvent)="formStatus($event)"></app-show-configuration>
-    </ng-container>
-    <!-- Advanced and security group configuration -->
-    <ng-container *ngFor="let g of advanceCategoriesGroup">
-      <ng-container *ngIf="g.category == category?.name+'Advanced'; else SecurityGroup">
+  <ng-container *ngIf="groups.length == 0 && advanceCategoriesGroup.length == 0; else group">
+    <div class="has-text-centered">
+      <small class="no-rec">No configuarion to update.</small>
```
praveen-garg, line 43: typo?

src/app/components/core/configuration-manager/configuration-group/configuration-group.component.html
```diff
@@ -38,33 +38,40 @@ <h6 class="is-6 is-small-font">
...
+    </div>
+  </ng-container>
+  <ng-template #group>
+    <div [ngClass]="{'highlight': from !== 'config-manager'}">
```
praveen-garg, line 47: I would like an inline comment for this !== condition

src/app/components/core/configuration-manager/configuration-group/configuration-group.component.html
```diff
@@ -38,33 +38,40 @@ <h6 class="is-6 is-small-font">
...
-          [selectedGroup]="selectedGroup" [fullConfiguration]="g.config" [from]="from"
-          (event)="getChangedAdvanceConfiguration($event)"
-          (formStatusEvent)="formStatus($event)"></app-show-configuration>
+          [from]="from" [selectedGroup]="selectedGroup" [fullConfiguration]="category.config"
+          (event)="getChangedConfiguration($event)" (formStatusEvent)="formStatus($event)"></app-show-configuration>
       </ng-container>
-      <ng-template #SecurityGroup>
-        <ng-container *ngIf="g.category == category?.name+'Security'; else pluginGroup">
+      <!-- Advanced and security group configuration -->
+      <ng-container *ngFor="let g of advanceCategoriesGroup">
```
praveen-garg, line 54: We should  name it dynamicCategories/  dynamicCategoriesGroup, dynamic can be runtime

src/app/components/core/configuration-manager/configuration-group/configuration-group.component.ts
```diff
@@ -239,21 +239,25 @@ export class ConfigurationGroupComponent implements AfterViewInit {
         e.group === 'Basic' ? acc.unshift(e) : acc.push(e);
         return acc;
       }, []);
+
+    // set advance as a first tab if no default config
```
praveen-garg, line 243: does not make any sense! Use case?

src/app/components/core/configuration-manager/configuration-group/configuration-group.component.ts
```diff
@@ -239,21 +239,25 @@ export class ConfigurationGroupComponent implements AfterViewInit {
...
+    if (this.groups.length == 0) {
+      this.selectedGroup = groups[0]?.group
```
praveen-garg, line 245: if length is zero then how it comes  groups[0]

src/app/components/core/filter/add-filter-wizard/add-filter-wizard.component.ts
```diff
@@ -341,6 +341,7 @@ export class AddFilterWizardComponent implements OnInit {
       this.validChildConfigurationForm = true;
       this.configurationData = null;
       this.pluginConfiguration = null;
+      pluginInfo.config = this.configurationControlService.getValidConfig(pluginInfo.config);
```
praveen-garg, line 344: among all `this.` move it to the top? and an inline comment explaining the case.

src/app/services/configuration-control.service.ts
```diff
@@ -538,4 +538,14 @@ export class ConfigurationControlService {
     return finalConfiguration;
   }
 
+  getValidConfig(config: any) {
+    // remove readonly property form the local configuration copy
+    Object.keys(config).forEach(key => {
+      if (config[key]?.readonly) {
+        delete config[key];
+      }
+    })
```
praveen-garg, line 547: Fine, but may be  return filtered

src/app/components/core/configuration-manager/configuration-group/configuration-group.component.ts
mshariq-nerd: when there is only readonly property in the default configuration, in that case no group for default configuration and on modal window dynamic property there first tab need to be set on first advance group

src/app/components/core/configuration-manager/configuration-group/configuration-group.component.ts
mshariq-nerd: global vs local group. Changed the name of local group to dynamicGroup for clear understanding

src/app/services/configuration-control.service.ts
mshariq-nerd: config is an object of properties objects and in there we need to remove the readonly key property, delete is used to remove that property on the same object and returning that object with remaining data.

monika-Dianomic (approved):

src/app/components/core/configuration-manager/configuration-group/configuration-group.component.html
```diff
@@ -38,33 +38,41 @@ <h6 class="is-6 is-small-font">
...
```
praveen-garg, line 55: I think you missed to change here dynamicCategoriesGroup

src/app/components/core/configuration-manager/configuration-group/configuration-group.component.ts
```diff
@@ -221,39 +221,43 @@ export class ConfigurationGroupComponent implements AfterViewInit {
...
+      dynamicGroups[i] = config;
     }
     else {
-      groups.push(config);
+      dynamicGroups.push(config);
     }
 
-    groups = groups
+    dynamicGroups = dynamicGroups
       .sort((a, b) => a.group.localeCompare(b.group))
```
praveen-garg, line 237: For better readability .sort should move to the line above

src/app/components/core/configuration-manager/configuration-group/configuration-group.component.ts
praveen-garg: I see what  you mean, Though it will be the rarest to see that; good to safeguard that

hemantgarg923 (approved):

PR #318, opened 2023-07-28
FOGL-7988: GUI support for control user

mshariq-nerd:


src/app/services/roles.service.ts
```diff
@@ -16,7 +16,21 @@ export class RolesService {
...
+  }
+
+  editPermissionBaseOnPageAccess(fromPage: string) {
+    if (fromPage == 'control-pipeline' && this.hasControlAccess()) {
+      return true;
+    } else if (fromPage !== 'control-pipeline' && this.hasEditPermissions()) {
+      return true;
+    }
+    return false;
   }
```
praveen-garg, line 34:
```
hasEditPermissions(page: string) { // if we cannot keep this name, then hasPageEditPermissions
    if (fromPage == 'control-pipeline') {
       return this.hasControlAccess();
     }  
   else {. // This else can also be  if fromPage != 'control-pipeline
       return this.hasEditPermissions();
     }
     return false;
}

```

src/app/services/roles.service.ts
```diff
@@ -16,7 +16,20 @@ export class RolesService {
...
+
+  public hasControlAccess(): boolean {
+    const roleId = Number(sessionStorage.getItem('roleId'));
+    return [appRoles.admin, appRoles.control, appRoles.anonymous].includes(roleId);
+  }
+
+  hasEditPermissionsOnPage(page: string) {
+    if (page == 'control-pipeline') {
+      return this.hasControlAccess();
+    } else {
```
praveen-garg, line 30: given the return, else is not required.

praveen-garg (approved):

hemantgarg923 (approved):

monika-Dianomic (approved):

PR #321, opened 2023-08-28
FOGL-7937: Playback buttons added on summary tab and spinning wheel added to represent graph API call

hemantgarg923:


src/app/components/core/asset-readings/readings-graph/readings-graph.component.html
```diff
@@ -117,7 +120,7 @@
         </div>
         <div class="columns">
           <div class="column">
-            <div id="tab-content">
+            <div id="tab-content" class="tab-content">
```
praveen-garg, line 123: it would be better to differentiate the id vs class names

praveen-garg (approved):

mshariq-nerd (approved):

monika-Dianomic (approved):

PR #335, opened 2023-10-04
Fixed responsiveness for mobile view

monika-Dianomic:


src/app/components/core/control-dispatcher/api-flow/add-edit-api-flow/add-edit-api-flow.component.css
```diff
@@ -83,3 +83,7 @@ span.is-small {
     font-size: 0.7em;
     color: gray;
 }
+
+.execute-dropdown {
```
praveen-garg, line 87: can not you use id="ed-dropdown"

src/app/components/core/support/support.component.css
```diff
@@ -10,3 +10,7 @@
 .is-hovered {
   cursor: pointer !important;
 }
+
+.help-pad {
```
praveen-garg, line 14: cannot we there mr-* class?

PR #341, opened 2023-10-12
FOGL-8156: South service & north service/task with name "add" redirected to details page instead of add wizard

hemantgarg923:


mshariq-nerd (approved):

praveen-garg (commented):
src/app/components/core/control-dispatcher/control-dispatcher.module.ts all checked?

hemantgarg923:
For control pipeline we are using id in route path and for script, acl and API entry point `/details` is added.

src/app/components/core/control-dispatcher/control-dispatcher.module.ts
```diff
@@ -94,7 +94,7 @@ const routes: Routes = [
     canActivate: [RolesGuard]
   },
   {
-    path: 'entry-point/:name',
+    path: 'entry-point/:name/details',
```
monika-Dianomic, line 97: I believe `\details` should also be added in the path of `ControlScheduleTaskDetailsComponent` in this module

mshariq-nerd (approved):

monika-Dianomic (approved):

PR #356, opened 2023-11-22
FOGL-8254: Application updated on Angular v16 

mshariq-nerd:


hemantgarg923:
Burger icon in navbar should appear if width <= 768px but it is appearing for width <= 1023px

PR #362, opened 2023-12-22
FOGL-8355: Update UI to use bucket properties as an object

mshariq-nerd:
Test with https://github.com/fledge-iot/fledge/pull/1252

### DIFF
--- a/src/app/services/configuration-control.service.ts
+++ b/src/app/services/configuration-control.service.ts
@@ -525,7 +525,7 @@ export class ConfigurationControlService {
     // make a copy of matched config items having changed values
     const matchedConfig = defaultConfig.filter(e1 => {
       if (changedConfiguration.hasOwnProperty(e1.key)) {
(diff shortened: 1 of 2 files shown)


src/app/services/configuration-control.service.ts
```diff
@@ -525,7 +525,7 @@ export class ConfigurationControlService {
     // make a copy of matched config items having changed values
     const matchedConfig = defaultConfig.filter(e1 => {
       if (changedConfiguration.hasOwnProperty(e1.key)) {
-        if (e1.type == 'JSON' || e1.type == 'bucket') {
+        if (e1.type == 'JSON') {
```
praveen-garg, line 528: This is the only place we had bucket type check?

praveen-garg (approved):

src/app/services/configuration-control.service.ts
mshariq-nerd: Yes

PR #373, opened 2024-01-30
Updated node v18 to fix e2e failing job

mshariq-nerd:


yarn.lock
```diff
@@ -2709,11 +2709,6 @@
   resolved "https://registry.yarnpkg.com/@types/node/-/node-14.18.63.tgz#1788fa8da838dbb5f9ea994b834278205db6ca2b"
   integrity sha512-fAtCfv4jJg+ExtXhvCkCqUKZ+4ok/JQk01qDKhL5BDDoS3AxKXhV5/MAVUZyQnSEd2GT92fkgZl0pz0Q0AzcIQ==
 
-"@types/node@^16.18.39":
```
praveen-garg, line 2712: What was the use of this?

e2e/po/app.skip.ts
```diff
@@ -34,7 +34,7 @@ export class SkipLogin {
   }
 
   getAppStatus() {
-    cy.get('#app > app-root > div > app-navbar > nav > div.navbar-menu .icon').should('be.visible')
```
praveen-garg, line 37: should we try something like cy.wait(2000)?

praveen-garg (approved):

yarn.lock
mshariq-nerd: This package is used to load in all type definitions when using typescript in node.

e2e/po/app.skip.ts
```diff
@@ -34,7 +34,8 @@ export class SkipLogin {
...
+    cy.wait(2000); // slight delay before checking app status
+    cy.get('app-navbar .navbar-menu .icon').should('be.visible')
```
praveen-garg, line 38: copy selector gives "#hover-link div span.icon" or just ""#hover-link div span" should be tried? or we can give a special ID to this and just get that ID.

hemantgarg923 (approved):

praveen-garg (approved):

PR #379, opened 2024-02-08
FOGL-8488: Fixed issue for development mode fledge and removed extra calls on clicking config icon/Configure link

monika-Dianomic:
**Done:** 
- [x] Handling done for development mode fledge on Additional Service modal and Additional Service list page
- [x] Removed extra API calls on clicking config icon and Configure link on list pages of different Services
- [x] Allow navigation to service modal by clicking service name appearing on hovering navbar status icon
- [x] Handling for pollingScheduleID when clicking from navbar status dropdown
- [x] Fixed issue - Enable/Disable service checkbox on modal is not showing correct status


src/app/components/core/developer/additional-services/additional-service-modal/additional-service-modal.component.ts
```diff
@@ -326,7 +326,7 @@ export class AdditionalServiceModalComponent {
     // enabling service takes time to get the updated state from API
     setTimeout(() => {
       this.toggleModal(false);             
-    }, 1000);
+    }, 3000);
     this.isServiceEnabled = true;
   }
```
mshariq-nerd, line 332: Why we need this timeout time increase?

src/app/components/core/developer/additional-services/additional-services-utils.service.ts
```diff
@@ -27,7 +27,7 @@ export class AdditionalServicesUtils {
               // enabling service takes time to get the updated state from API
               setTimeout(() => {
                 this.router.navigate(['/developer/options/additional-services']);             
-              }, 2000);
+              }, 3000);
```
mshariq-nerd, line 30: As above

monika-Dianomic: Reason is already mentioned there

src/app/components/core/control-dispatcher/api-flow/api-flow.component.ts
```diff
@@ -44,6 +45,7 @@ export class APIFlowComponent implements OnInit {
     public reenableButton = new EventEmitter<boolean>(false);
 
     showConfigureModal: boolean = false;
+    serviceInfo;
```
mshariq-nerd, line 48: What this variable hold? Any type definition?

src/app/components/core/control-dispatcher/list-control-dispatcher/acl-list/acl-list.component.ts
```diff
@@ -27,6 +27,7 @@ export class AclListComponent implements OnInit {
 
   public reenableButton = new EventEmitter<boolean>(false);
   showConfigureModal: boolean = false;
+  serviceInfo;
```
mshariq-nerd, line 30: As above

src/app/components/core/control-dispatcher/add-dispatcher-service/add-dispatcher-service.component.ts
```diff
@@ -16,11 +15,11 @@ export class AddDispatcherServiceComponent implements OnInit {
   dispatcherServiceInstalled;
   dispatcherServiceAdded;
   dispatcherServiceEnabled;
+  dispatcherServiceName: string = '';
   showConfigureModal: boolean = false;
```
mshariq-nerd, line 19: Just for FYI. There is no need to explicitly telling the type of the variable if you are passing the default value as typescript automatically convert the type based on the default value. Like here no need to give type boolean as you are passing default that is a boolean itself.
`showConfigureModal = false;` // typecast automatically into boolean
dispatcherServiceName = ''; // typecast automatically into string

src/app/components/core/developer/additional-services/additional-service-modal/additional-service-modal.component.ts
```diff
@@ -70,18 +75,31 @@ export class AdditionalServiceModalComponent {
...
   ngOnInit() { }
 
-  getServiceInfo(serviceInfo, availableServicePkgs, pollingScheduleID) {
+  getServiceInfo(serviceInfo, pollingScheduleID, from = null) {
+    this.navigateFromParent = from;
     this.serviceName = serviceInfo.name ? serviceInfo.name : '';
-    this.isServiceEnabled = ["shutdown", "disabled", "installed", ""].includes(serviceInfo.state) ? false : true;
-    this.isServiceAvailable = serviceInfo.added;
+    this.isServiceEnabled = serviceInfo.isEnabled ? (serviceInfo.isEnabled === 'true' || serviceInfo.isEnabled === true) : false;
+    this.isServiceAvailable = serviceInfo.added ? (serviceInfo.added === 'true' || serviceInfo.added === true) : false;
```
mshariq-nerd, line 98: `(serviceInfo.added === 'true' || serviceInfo.added === true) `
It means serviceInfo.added can hold both type of values string & boolean? This is not correct it should either be boolean or string and should be handled while creating the serviceInfo object. 
same is true for 
`serviceInfo.isEnabled === 'true' || serviceInfo.isEnabled === true`

src/app/components/core/developer/additional-services/additional-service-modal/additional-service-modal.component.ts
```diff
@@ -70,18 +75,31 @@ export class AdditionalServiceModalComponent {
...
     this.serviceProcessName = serviceInfo.process;
     this.serviceType = serviceInfo.type;
     this.packageName = serviceInfo.package;
-    this.availableServices = availableServicePkgs;  
+    this.isInstalled =  serviceInfo.isInstalled ? (serviceInfo.isInstalled === 'true' || serviceInfo.isInstalled === true) : false;
```
mshariq-nerd, line 102: As above

src/app/components/layout/navbar/navbar.component.html
```diff
@@ -244,7 +244,7 @@
                                 <i class="fa fa-sm fa-circle child-icon"></i>
                               </span>
                               <ng-container
-                                *ngIf="['northbound', 'southbound'].includes(s.type.toLowerCase()); else elseServiceTemplate">
+                                *ngIf="['northbound', 'southbound', 'dispatcher', 'notification', 'management', 'bucketstorage'].includes(s.type.toLowerCase()); else elseServiceTemplate">
```
mshariq-nerd, line 247: better to move `['northbound', 'southbound', 'dispatcher', 'notification', 'management', 'bucketstorage'].` array in ts. May be in utils.ts?

src/app/components/layout/navbar/navbar.component.ts
```diff
@@ -426,9 +427,70 @@ export class NavbarComponent implements OnInit, AfterViewInit, OnDestroy {
...
-    let routePath = service.type === 'Northbound' ? 'north' : 'south';
-    this.router.navigate([routePath, service.name, 'details'])
+  public showConfigWithPollingSchedule(serviceInfo): void {
+    this.ngProgress.start();
+    this.schedulesService.getSchedules().
+      subscribe(
+        (data: any) => {
+          const pollingScheduleID = data.schedules.find(s => s.processName === 'manage')?.id;
+          serviceInfo['pollingScheduleID'] = pollingScheduleID;
+          this.router.navigate(['/developer/options/additional-services/config'], { queryParams: { ...serviceInfo },  skipLocationChange: true });
```
mshariq-nerd, line 437: Missing `this.ngProgress.done();` in this code block?

src/app/components/core/developer/additional-services/additional-service-modal/additional-service-modal.component.ts
```diff
@@ -70,18 +75,31 @@ export class AdditionalServiceModalComponent {
...
+    private additionalServicesUtils: AdditionalServicesUtils,
+    public rolesService: RolesService) {
+      this.activatedRoute.queryParams.subscribe(params => {
+        if (params?.name) {
+            this.fromNavbar = true;
+            this.getServiceInfo(params, params?.pollingScheduleID);
+            // Wait to get html page loaded
+            setTimeout(() => {
+              this.toggleModal(true);
+            }, 500);
```
mshariq-nerd, line 87: We should carefully use the set timeout because the time given is not a guarantee that page will be available in that time.

src/app/components/core/developer/additional-services/additional-service-modal/additional-service-modal.component.ts
```diff
@@ -298,21 +321,29 @@ export class AdditionalServiceModalComponent {
     if (name != null) {
       serviceName = name;
     }
-    this.toggleModal(false, {service: serviceName, state: 'enable', isCancelEvent: false});
+    this.additionalServicesUtils.enableService(serviceName, this.fromNavbar);
+    // added some delay to close the modal because 
+    // enabling service takes time to get the updated state from API
+    setTimeout(() => {
+      this.toggleModal(false);
```
mshariq-nerd, line 328: We should carefully use the set timeout because the time given is not a guarantee that page will be available in that estimated time.

src/app/components/core/developer/additional-services/additional-services-utils.service.ts
mshariq-nerd: How we can ensure that it will be available 3000 ms as well?

monika-Dianomic: I've tested it multiple times with this timestamp and it always passed but If you have better idea then let me know how can we ensure about the exact time?

src/app/components/layout/navbar/navbar.component.html
monika-Dianomic: This is being used at only one place (i.e. only here), so is this a better idea to move this array in utils.ts? What will be the benefit of it?

src/app/components/core/developer/additional-services/additional-service-modal/additional-service-modal.component.ts
monika-Dianomic: This component is being called via route (not via child component) in the navbar. There is a reason behind it (I'll let you know) and from navbar this `serviceInfo` is being passed via `queryParams`. As you know, `queryParams` automatically converted into string. So I've to do it like above to get the values of variables.

src/app/components/core/control-dispatcher/add-dispatcher-service/add-dispatcher-service.component.ts
monika-Dianomic: Thanks for the clarification but I already know about it. I believe that there is no harm in that form also but if you suggest to not write in this way then I'll take care of it.

src/app/components/layout/navbar/navbar.component.ts
```diff
@@ -426,9 +427,71 @@ export class NavbarComponent implements OnInit, AfterViewInit, OnDestroy {
...
+        this.router.navigate(['north', service.name, 'details']);
+        break;
+      case 'Southbound':
+        this.router.navigate(['south', service.name, 'details']);
+        break;
+      case 'BucketStorage':
+        serviceInfo['process'] = 'bucket';
+        serviceInfo['type'] = 'BucketStorage';
+        serviceInfo['package'] = 'fledge-service-bucket';
+        this.router.navigate(['/developer/options/additional-services/config'], { queryParams: { ...serviceInfo },  skipLocationChange: true });
```
mshariq-nerd, line 473: Sending big object in the queryParams not looks a good idea to me
```
{
    "name": "N1",
    "isEnabled": "true",
    "added": "true",
    "isInstalled": "true",
    "process": "notification",
    "type": "Notification",
    "package": "fledge-service-notification"
}
```

And why we need to do `skipLocationChange: true`? To hide the object in the queryParam? 
By doing this the path `/developer/options/additional-services/config` will not be visible in the browser Address bar ever.

There must be better way to do this.

mshariq-nerd (changes requested):

src/app/components/core/developer/additional-services/additional-services-utils.service.ts
mshariq-nerd: Think about the low end device like RPi, Have you tested on that? We should look for better observability and see how to trigger the event when previous call in the stack completed to notify for next action.

src/app/components/layout/navbar/navbar.component.ts
```diff
@@ -49,6 +49,7 @@ export class NavbarComponent implements OnInit, AfterViewInit, OnDestroy {
   viewPort: any = '';
   public showSpinner = false;
   isManualRefresh = false;
+  servicesToShow = ['northbound', 'southbound', 'dispatcher', 'notification', 'management', 'bucketstorage'];
```
mshariq-nerd, line 52: rename servicesToShow => externalServices

src/app/components/core/developer/additional-services/additional-service-modal/additional-service-modal.component.ts
```diff
@@ -77,29 +77,31 @@ export class AdditionalServiceModalComponent {
     private configurationControlService: ConfigurationControlService,
     private additionalServicesUtils: AdditionalServicesUtils,
     public rolesService: RolesService) {
-      this.activatedRoute.queryParams.subscribe(params => {
-        if (params?.name) {
+      this.activatedRoute.paramMap
+      .pipe(map(() => window.history.state)).subscribe(res=>{
```
mshariq-nerd, line 81: `map(() => window.history.state)` is this required?

src/app/components/layout/navbar/navbar.component.ts
monika-Dianomic: 'northbound' and 'southbound' services are not externalServices. Due to which I didn't named it as `externalServices `.

src/app/components/core/developer/additional-services/additional-service-modal/additional-service-modal.component.ts
monika-Dianomic: Yes, it is required to fetch passed object from navbar component.

mshariq-nerd (dismissed):

mshariq-nerd (approved):

PR #386, opened 2024-02-27
Some code refactoring and removal of extra calls on Additional Services page

monika-Dianomic:


src/app/components/common/node-editor/node-editor.component.ts
```diff
@@ -668,9 +668,9 @@ export class NodeEditorComponent implements OnInit {
     }
   }
 
-  deleteService() {
-    this.servicesApiService.deleteService(this.deleteServiceName)
-      .subscribe((data: any) => {
+  async deleteService() {
+    await this.servicesApiService.deleteService(this.deleteServiceName)
+      .then((data: any) => {
```
mshariq-nerd, line 673: await with then?  mixing 2 same things together. 
> Async/Await is a syntactic sugar for promises, a wrapper making the code execute more synchronously

> In Promise, then method called when the promise becomes fulfilled. 

We should either use async/await or Promise.then not mixing the same things. Also not sure do we really need this change?

mshariq-nerd (changes requested):

src/app/components/core/developer/additional-services/additional-service-modal/additional-service-modal.component.ts
```diff
@@ -177,8 +170,8 @@ export class AdditionalServiceModalComponent {
     if (!installationState) {
       this.ngProgress.start();
     }
-    this.servicesApiService.addService(payload)
-      .subscribe(
+    await this.servicesApiService.addService(payload)
+      .then(
```
mshariq-nerd, line 174: As above async/await vs then

hemantgarg923 (approved):

PR #398, opened 2024-03-15
FOGL-8306: Alerts stuff (patch3)

MonikaSharma06:


src/app/components/core/system-alert/system-alert.component.ts
```diff
@@ -15,15 +17,23 @@ export class SystemAlertComponent {
   expectedButtonLabels = ['Show Logs', 'Upgrade'];
   sortByKey = 'time';
   showSpinner = true;
+  isManualRefresh = false;
```
hemantgarg923, line 20: May be we could use the variable name `isAlive` instead of `isManualRefresh`, which is the standard name for checking ping state throughout the application.

hemantgarg923 (dismissed):

src/app/components/core/system-alert/system-alert.component.ts
MonikaSharma06: We use variable name `isAlive` on the pages wherever we are doing auto refresh call e.g.- South, North etc. `isAlive` indicates that we should keep auto refresh call alive or not.
But here I need to only check that `Auto ping` is set Or `Manual`.
Due to above reason, I think `isManualRefresh` is well suited here.

praveen-garg: I think Hemant meant that We don't this variable if we have access to pingInfo object, and can use pingInfo.isAlive along with alertsCount

src/app/components/core/system-alert/system-alert.component.ts
```diff
@@ -15,15 +17,23 @@ export class SystemAlertComponent {
...
+    private ping: PingService,
     private systemAlertService: SystemAlertService,
     public ngProgress: ProgressBarService,
     private rolesService: RolesService) {
-  }
+      this.ping.pingIntervalChanged
+      .pipe(takeUntil(this.destroy$))
+      .subscribe((pingTime: number) => {
+        this.isManualRefresh = pingTime === -1 ? true : false;
+      });
```
mshariq-nerd, line 35: I think we can avoid this extra subscription of pingIntervalChanged here. We always have a active subscription for ping interval check in navbar. We can pass the information of isManualRefresh from the navbar as an Input() and use that information in ngOnChanges e.g. 
```
 if (changes?.alertsCount) {
      this.alertsCount = changes?.alertsCount?.currentValue;
    }
    if (changes?.isManualRefresh) {
      this.isManualRefresh = changes?.isManualRefresh?.currentValue;
    }

mshariq-nerd (approved):

hemantgarg923 (approved):

PR #405, opened 2024-04-04
FOGL-8579_GUI: Updation as per backend changes

MonikaSharma06:
Required to sync with https://github.com/fledge-iot/fledge/pull/1325 changes.

### DIFF
--- a/src/app/components/core/system-alert/system-alert.component.ts
+++ b/src/app/components/core/system-alert/system-alert.component.ts
@@ -101,8 +101,8 @@ export class SystemAlertComponent {
         setTimeout(() => {
           this.hideLoadingSpinner();
         }, 500);
-        data.alerts.forEach(alert => {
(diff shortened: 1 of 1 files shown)


src/app/components/core/system-alert/system-alert.component.ts
```diff
@@ -249,7 +249,7 @@ export class SystemAlertComponent {
     if (message.includes('restarted')) {
       return "Show Logs";
     }
-    if (message.includes('upgrade')) {
+    if (message.includes('updates') || message.includes('update')) {
```
praveen-garg, line 252: We should first check based on key. 

I understand the problem that service restart does not have a defined key and that still need to rely on text parsing.

praveen-garg (changes requested):

praveen-garg (approved):

mshariq-nerd (approved):

PR #408, opened 2024-04-10
FOGL-8509: changes to make all side menu options visible

MonikaSharma06:


praveen-garg (changes requested):

src/app/components/layout/side-menu/side-menu.component.css
```diff
@@ -1,11 +1,11 @@
 .menu-list {
   background-color: white !important;
-  min-height: 100vh;
-  margin-top: 0.25%;
+  width: 100%;
```
praveen-garg, line 3: Why is this required?

MonikaSharma06: this makes slightly change in width. But not very much required. I'll remove it.

praveen-garg (approved):

hemantgarg923 (approved):

PR #412, opened 2024-04-22
FOGL-8615: Notification flow editor

MonikaSharma06:
### **TODO**

**Notifications Editor:** 
- [ ] Show logs for notify service name + that selected instance name; not only the selected notification instance
- [ ] Add node should be disabled not hidden, when service is not installed etc.
- [ ] At initial stage, Save button should be disable at config quick view
- [ ] Alignment issue of notify plugin name

**Common:**
- [ ] Change Add (empty) node as 2px dotted instead of solid
- [ ] Add node should be hidden in case of View User
- [ ] In right drawer, make Auto Refresh label font-size 400

**Other Components:**
- [ ] Separate service to get installed/available services object has been added, need to remove this code from additional services component and use it from the common service
- [ ] NotificationServiceWarningComponent can be used as common warning component for dispatcher service also


src/app/components/core/notifications/notification-service-config/notification-service-config.component.html
```diff
@@ -0,0 +1,5 @@
+<a *ngIf="rolesService.hasControlAccess()" class="button" (click)="openConfigureModal()" id="config-icon">
+    <span class="icon">
+        <i class="bi bi-gear-wide-connected bi-sm" aria-hidden="true" title="Notification Service Configuration"></i>
+    </span>
+</a>
```
mshariq-nerd, line 5: We already have an identical component `dispatcher-service-config.component`. Better to make that generic component instead separate for each service.

mshariq-nerd (changes requested):

praveen-garg (dismissed):
I assume there is no regression in

1) Flow Editor ON- South and North
2) Flow Editor OFF - Notifications pages
3) Dispatcher config pages

src/app/components/common/node-editor/node-editor.component.ts
```diff
@@ -79,6 +79,18 @@ export class NodeEditorComponent implements OnInit {
...
+  rulePluginChangedConfig: any;
+  ruleCategory: any;
+  deliveryCategory: any;
+  deliveryPluginChangedConfig: any;
+  ruleConfiguration: any;
+  deliveryConfiguration: any;
+  showConfigureModal = false;
+  isServiceAvailable = false;
+  serviceInfo: {};
+  btnText = '';
```
mshariq-nerd, line 92: Can't we avoid those many variables for managing the notification instance. We should look for the better data structure to avoid all these variables.

src/app/components/common/node-editor/node-editor.component.html
```diff
@@ -68,29 +82,77 @@
...
+    <ng-container *ngIf="showLogs && from !== 'notifications'">
       <app-system-log [sourceName]="serviceName"></app-system-log>
     </ng-container>
+
+    <!-- Notification Log in Quickview -->
+    <ng-container *ngIf="showLogs && from === 'notifications'">
+      <app-notification-log [sourceName]="serviceName"></app-notification-log>
+    </ng-container>
+
+    <!-- Notification Instance Configuration in Quickview -->
```
hemantgarg923, line 94: Better to move these changes in a separate component

src/app/components/common/node-editor/custom-notification-node/custom-notification-node.component.ts
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  ngOnChanges(): void {
+    this.nodeId = this.data.id;
+    if (this.data.label === 'Notification') {
+      this.elRef.nativeElement.style.borderColor = "#0E9BD8";
+      this.isNotificationNode = true;    
+      if ('notifications' == 'notifications') {
+        if (!isEmpty(this.data.controls)) {
+          this.notification.name = this.data.controls.nameControl['name'];
```
hemantgarg923, line 80: Better to use data.controls directly in html instead of creating separate object and initialise its properties with incoming data

MonikaSharma06: We are passing this notification data to `flowEditorService.showItemsInQuickview` and it is of Notification type i.e.-


```
 {
    name: string
    rule: string
    channel: number
    notificationType: string
    retriggerTime: number
    enable: boolean
  }
```

If I pass it as it is i.e. `this.data.controls` then I'll have to access all controls as `this.notification.nameControl['name']` etc. in node editor file and it can't be typecast as `Notification`. So I think it is better to use an notification object here.

src/app/components/common/node-editor/node-editor.component.ts
MonikaSharma06: I've removed one variable i.e. `showConfigureModal`. All other variables seems required, which I can't remove.

src/app/components/common/node-editor/node-editor.component.html
MonikaSharma06: We can handle it in the next patch PR.

mshariq-nerd (approved):

PR #419, opened 2024-05-20
FOGL-8715: Fixed undefined connection issue while changing filter pipeline

mshariq-nerd:


praveen-garg (approved):

src/app/components/common/node-editor/insert-node/index.ts
```diff
@@ -54,12 +54,10 @@ export function insertableNodes<S extends Schemes>(
...
       if (view && node.label !== "South" && node.label !== "Storage" && node.label !== "North") {
-        const intersectedConnections = checkIntersection(view.position, node, cons);   
-
-        for(let id of intersectedConnections){
+        const intersectedConnections = checkIntersection(view.position, node, cons);
+        for (let id of intersectedConnections) {
           const exist = editor.getConnection(id);
-
-          if (exist.source !== node.id && exist.target !== node.id) {
+          if (exist && (exist.source !== node.id && exist.target !== node.id)) {
```
praveen-garg, line 60: For curiosity,  We could have had (exist?.source !== node.id && exist?.target !== node.id)?

mshariq-nerd: Yes, that can be another way of checking using the optional chaining operator.
Since undefined are both falsy values , the property after the && operator is only accessed if the precedent it not null or undefined.

praveen-garg: Yeah, the check you added is more clean and easily readable.

hemantgarg923 (approved):

MonikaSharma06 (approved):

PR #422, opened 2024-05-28
FOGL-8684: Added support for kvlist enumeration config item type

MonikaSharma06:


src/app/components/core/configuration-manager/kv-list-type-configuration/kv-list-type-configuration.component.ts
```diff
@@ -25,6 +25,9 @@ export class KvListTypeConfigurationComponent implements OnInit {
 
   ngOnInit() {
     this.form = this.rootFormGroup.control;
+    if (this.configuration.items == 'enumeration' && this.configuration.options && typeof(this.configuration.options) == 'string'){
+      this.configuration.options = JSON.parse(this.configuration.options)
+    }
```
mshariq-nerd, line 30: Do we really need this check? To me API always returns options as a JSON array so no need of JSON.parse here. 
If backend is returning options in a string then it's a backed issue and we should log that?

hemantgarg923 (approved):

mshariq-nerd (approved):

PR #423, opened 2024-05-29
FOGL-8773: Added health icon for north service node

MonikaSharma06:


mshariq-nerd (changes requested):
Need to change the updateNode function in editor.ts as well to reflect the node status icon based on the service state change on checkbox selection to enable/disable node. 
```
After line 452
statusControl.status = task.status;
```
 and in north-task.ts an optional property status
 ```
export interface NorthTask {
...
...
...
status?: string;
}
```

hemantgarg923 (dismissed):

mshariq-nerd (approved):

PR #424, opened 2024-06-03
FOGL-8237:  Provide facility for reading data from JSON file in GUI for JSON type configuration item field

mshariq-nerd:


praveen-garg:
When I upload a file it should appear in pretty format (with proper indents); As it appears after save and reload.

src/app/services/configuration-control.service.ts
```diff
@@ -581,9 +581,13 @@ export class ConfigurationControlService {
       if (changedConfiguration.hasOwnProperty(e1.key)) {
```
MonikaSharma06, line 581: rather than using `e1`, a better variable name should be used here

src/app/components/core/configuration-manager/show-configuration/show-configuration.component.html
```diff
@@ -39,11 +39,23 @@
...
-                      [options]="setCodeMirrorOption(configuration)" [autoFocus]="true"
-                      [jsonValue]="form.controls[configuration.key]">
-                    </ngx-codemirror>
+                    <ng-container *ngSwitchCase="'JSON'">
+                      <ngx-codemirror [formControlName]="configuration.key" #jsonEditor
+                        [ngClass]="{'readonly-editor': form.controls[configuration.key]?.status === 'DISABLED' }"
+                        [options]="setCodeMirrorOption(configuration)" [autoFocus]="true"
+                        [jsonValue]="form.controls[configuration.key]">
+                      </ngx-codemirror>
+                      <p class="help" *ngIf="configuration?.fileName">
```
hemantgarg923, line 48: This element could be removed as no file is being created on server, so filename property does not exist for JSON type config item.

hemantgarg923 (approved):

PR #425, opened 2024-06-04
FOGL-8783: Removed extra sytem/notifications log API calls in flow editor

MonikaSharma06:


src/app/components/core/logs/notification-log/notification-log.component.ts
```diff
@@ -58,6 +58,12 @@ export class NotificationLogComponent implements OnInit, OnDestroy {
     this.subscription = interval(this.refreshInterval)
       .pipe(takeWhile(() => this.isAlive), takeUntil(this.destroy$)) // only fires when component is alive
       .subscribe(() => {
+        if (this.sourceName) {
+          const quickView = <HTMLDivElement>document.getElementById('quickviewDefault');
+          if (!quickView.classList.contains('is-active')) {
+            return;
+          }
+        }
```
mshariq-nerd, line 66: By this check we are only returning from the Interval Observable but the Interval() Observable will stay active in the background. To stop the Interval Observable completely on page close we need to see how we can stop that Observable. To do that we need to unsubscribe the this.subscription.

mshariq-nerd (changes requested):

src/app/components/core/logs/system-log/system-log.component.ts
```diff
@@ -67,6 +67,12 @@ export class SystemLogComponent implements OnInit, OnDestroy {
     this.subscription = interval(this.refreshInterval)
       .pipe(takeWhile(() => this.isAlive), takeUntil(this.destroy$)) // only fires when component is alive
       .subscribe(() => {
+        if (this.sourceName) {
+          const quickView = <HTMLDivElement>document.getElementById('quickviewDefault');
+          if (!quickView.classList.contains('is-active')) {
+            return;
+          }
+        }
```
mshariq-nerd, line 75: As above

mshariq-nerd (approved):

hemantgarg923 (approved):

PR #429, opened 2024-06-21
FOGL-8862: Hide Support & 'Backup and Restore' tabs for all other role users except admin role.

mshariq-nerd:


src/app/services/roles.service.ts
```diff
@@ -10,6 +10,16 @@ export class RolesService {
   private roleNames = [{ roleId: 1, name: "Administrator" }, { roleId: 2, name: "Editor" }, { roleId: 3, name: "Viewer" }, { roleId: 4, name: "Data Viewer" }, { roleId: 5, name: "Control" }];
 
   constructor() { }
+
+  /**
+   * To check if user is admin or anonymous to access backup/restore
+   * and support tabs in sidebar
+   */
+  public hasAccessPermissions(): boolean {
```
praveen-garg, line 18: hasAdminPermissionsOrAnonymousAllowed?

src/app/services/roles.service.ts
```diff
@@ -10,6 +10,16 @@ export class RolesService {
   private roleNames = [{ roleId: 1, name: "Administrator" }, { roleId: 2, name: "Editor" }, { roleId: 3, name: "Viewer" }, { roleId: 4, name: "Data Viewer" }, { roleId: 5, name: "Control" }];
 
   constructor() { }
+
+  /**
+   * To check if user is admin or anonymous to access backup/restore
```
praveen-garg, line 15: Remove backup & restore / support text

src/app/guards/index.ts
```diff
@@ -3,3 +3,5 @@ export * from './admin.guard';
 export * from './auth-required.guard';
 export * from './roles.gurad';
```
praveen-garg, line 4: Fixed roles guard as well?

src/app/guards/index.ts
```diff
@@ -3,3 +3,5 @@ export * from './admin.guard';
...
 export * from './data-view-role.guard'
```
praveen-garg, line 5: When we roles guard, why do we need another as data view guard.

For now its okay to keep to avoid last minute issues, but add a Jira to simplify role based access guards.

hemantgarg923 (approved):

PR #430, opened 2024-06-24
FOGL-8825: Removed extra GET notifications API call on Notification page (Flow editor)

MonikaSharma06:
**Done:**
- [x] FOGL-8825 (Removed extra notification API call)
- [x] Show logs for notif service name + that selected notif instance name only
- [x] Add node should be disabled not hidden
- [x] Fixed Node size issue

### DIFF
--- a/src/app/components/common/node-editor/custom-node/custom-node.component.html
+++ b/src/app/components/common/node-editor/custom-node/custom-node.component.html
@@ -172,7 +172,7 @@
     </div>
   </ng-container>
   <ng-template #addServiceTemplate>
-    <div [ngClass]="{'disabled-node': !rolesService.hasEditPermissions()}" [ngClass]="from == 'north' ? 'add-task' : 'add-service'">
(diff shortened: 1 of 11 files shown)


src/app/components/common/node-editor/custom-node/custom-node.component.html
```diff
@@ -172,8 +172,8 @@
     </div>
   </ng-container>
   <ng-template #addServiceTemplate>
-    <div class="add-service">
-      <span class="has-tooltip-bottom has-tooltip-arrow tooltip is-hovered"
+    <div [ngClass]="{'disabled-node': !rolesService.hasEditPermissions()}" [ngClass]="from == 'north' ? 'add-task' : 'add-service'">
```
hemantgarg923, line 175: Using `ngClass` directive twice on same element actually overrides the first occurrence. To apply multiple classes with `ngClass` we could simply use comma separated syntax.
`[ngClass]="{'disabled-node': !rolesService.hasEditPermissions(), 'add-task': from=='north', 'add-service': from!='north'}"`

mshariq-nerd (approved):

PR #437, opened 2024-07-09
FOGL-8841:  GUI support for permission properties in the configuration category

mshariq-nerd:


src/app/services/roles.service.ts
```diff
@@ -33,6 +33,17 @@ export class RolesService {
     return [appRoles.admin, appRoles.control, appRoles.anonymous].includes(roleId);
   }
 
+  public hadEditPermissionsOnConfigItem(allowedRoles: string[] = []) {
+    try {
+      const roleId = Number(sessionStorage.getItem('roleId'));
+      const roleName = appRoles[roleId];
```
praveen-garg, line 39: I assume _anonymous_ will not be an issue.

mshariq-nerd: yes, that is fine. Filed will be editable for anonymous(authenticationOptional:true).

src/app/components/core/configuration-manager/kv-list-type-configuration/kv-list-type-configuration.component.html
```diff
@@ -4,55 +4,84 @@
...
+            </ng-template>
           </ng-container>
           <ng-template #objectTemplate>
             <div class="column pt-5 is-11">
               <div class="field is-horizontal">
                 <div class="field-label has-text-left">
                   <label class="label is-inline-block">Key</label>
-                  <span class="ml-1">
-                    <i class="bi bi-xs bi-info-circle info-icon" aria-hidden="true" [title]="configuration?.description"></i>
+                  <span class="ml-1" *ngIf="rolesService.hasAccessPermission(configuration?.permissions)">
```
hemantgarg923, line 48: Do we want to hide kvlist config item description from non-privileged user?

hemantgarg923 (approved):

MonikaSharma06 (approved):

PR #438, opened 2024-07-10
FOGL-8679: Fixed plugin version tooltip issue on south and north page for flow editor

hemantgarg923:


praveen-garg (commented):
Explanation of true to false switching?

praveen-garg (commented):
Filter
Rule/Notify all okay?

hemantgarg923:
> Filter Rule/Notify all okay?

Currently version tooltip is available only on south and north services/tasks

hemantgarg923:
> Explanation of true to false switching?

Here true/false is used for caching. Previously the call was cached, which should not be the case so switched to cached=false.

mshariq-nerd (approved):

PR #444, opened 2024-07-17
FOGL-8694: added context menu with delete option for connection deletion

mshariq-nerd:
To delete a connection, user can right click on the connection and delete it using the delete option context menu.

### DIFF
--- a/src/app/components/common/node-editor/editor.ts
+++ b/src/app/components/common/node-editor/editor.ts
@@ -113,11 +113,9 @@ export async function createEditor(container: HTMLElement, injector: Injector, f
           label: 'Delete',
           key: 'delete',
           async handler() {
-            if ('source' in context && 'target' in context) {
-              // connection
-              const connectionId = context.id
-              await editor.removeConnection(connectionId)
-            }
+            // connection
+            const connectionId = context.id
+            await editor.removeConnection(connectionId)
           }
         }
         return {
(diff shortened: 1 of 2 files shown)


src/app/components/common/node-editor/editor.ts
```diff
@@ -105,16 +105,30 @@ export async function createEditor(container: HTMLElement, injector: Injector, f
...
-              }
+          list: []
+        }
+      }
+      if ('source' in context && 'target' in context) {
+        const deleteItem = {
+          label: 'Delete',
+          key: 'delete',
+          async handler() {
+            if ('source' in context && 'target' in context) {
```
praveen-garg, line 116: Do We need this if again?

src/app/components/common/node-editor/editor.ts
```diff
@@ -105,16 +105,30 @@ export async function createEditor(container: HTMLElement, injector: Injector, f
       if (context === 'root') {
         return {
           searchBar: false,
-          list: [
-            {
-              label: 'Filter', key: '1', handler: () => {
-                let filter = new Filter(socket, { pluginName: '', enabled: 'false', filterName: 'Filter' });
```
praveen-garg, line 111: Is it safe to remove? 

What was this being used for?

mshariq-nerd: It was not is used. It was for providing the add filter context menu on the canvas but we did not add that feature.

hemantgarg923 (approved):

MonikaSharma06 (approved):

PR #450, opened 2024-08-19
FOGL-8984: Flow Editor: Routing doesn’t change if entity exists in the URL is no longer available (after resetting fledge/foglamp)

mshariq-nerd:


src/app/components/common/node-editor/node-editor.component.ts
```diff
@@ -316,26 +316,28 @@ export class NodeEditorComponent implements OnInit {
       let isServiceExist = true;
       if (this.initialApiCallsStack.length > 0) {
         this.ngProgress.start();
+        let retries = 4; // Retries
         forkJoin(this.initialApiCallsStack)
           .pipe(mergeMap(res => {
-            // Retry GET tasks call when task as a service created. Service takes time to populate in the GET tasks response
+            // Retry GET tasks call (retries + 1) time when task as a service created, It takes time to populate in the GET tasks response
```
MonikaSharma06, line 322: Refreshing the north detail page, after deleting the task, taking more time than expected. 6 API calls are there of GET | north API call. Can we resolve this issue?

mshariq-nerd: @MonikaSharma06, We have a retry mechanism of the GET north api call because when we create a north service, it takes time at backend side to create and populate that service in the response. 
So we have to check if the service name exist in the response that is there in the details page route and retry 5 times with 1500ms delay in every subsequent call. That is why you are seeing that much delay as it waits to finish all the retries before exit.
This delay will happen when user is on the details page of the task service and in the background you have deleted that service/task. 
On page hard refresh, because in the task details route the name of the task exist so retry mechanism try to find the name in tasks GET response 5 times and after all the attempts it exit and user moved on the list page.

MonikaSharma06 (approved):

praveen-garg (approved):

PR #451, opened 2024-08-20
FOGL-9033: Changed properties value sent in payload for object type list & kvlist

hemantgarg923:


praveen-garg:
If this fixes the problem. Let's merge it. 

Code smell and reusing validation part can be done in a patch PR.

cyberwalk3r (approved):

MonikaSharma06 (approved):

PR #454, opened 2024-08-29
Some css fixes

MonikaSharma06:
- [x] Remove border radius from all select input
- [x] Fixed alignment of fields in view mode
- [x] Made `height: auto` for codemirror editor
- [x] Changed opacity of configuration tree icon

### DIFF
--- a/src/app/components/core/configuration-manager/static-configuration/static-configuration.component.html
+++ b/src/app/components/core/configuration-manager/static-configuration/static-configuration.component.html
@@ -36,7 +36,7 @@
             class="readonly-editor" [disabled]="true">
           </ngx-codemirror>
         </ng-container>
-        <span *ngSwitchDefault class="help p-1 m-0">{{configuration.value ? configuration.value : "None"}}</span>
+        <span *ngSwitchDefault class="help">{{configuration.value ? configuration.value : "None"}}</span>
       </div>
     </div>
   </div>
--- a/src/styles.css
+++ b/src/styles.css
@@ -468,6 +468,7 @@ span.tooltip {
 
 .CodeMirror {
   height: auto;
+  max-height: 300px;
 }
 
 /*****************/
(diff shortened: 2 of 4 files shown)


src/styles.css
```diff
@@ -462,4 +462,13 @@ span.tooltip {
   font-size: 0.9rem !important;
 }
 
+.select select {
+  border-radius: 0px;
```
praveen-garg, line 466: On other controls like input, Do we have 0px or 2px?

src/app/components/core/configuration-manager/static-configuration/static-configuration.component.html
```diff
@@ -36,7 +36,7 @@
             class="readonly-editor" [disabled]="true">
           </ngx-codemirror>
         </ng-container>
-        <span *ngSwitchDefault class="help p-1 m-0">{{configuration.value ? configuration.value : "None"}}</span>
```
praveen-garg, line 39: What was the issue with p-1 m-0? Just want to make sure, We know why are we removing this.

src/styles.css
MonikaSharma06: It is 2px and 4px on different pages. Should I change and make all of these consistent?

src/app/components/core/configuration-manager/static-configuration/static-configuration.component.html
MonikaSharma06: This was only for one type of control and it was causing alignment issue on configuration page. I don't see any need of it so I removed.

src/styles.css
praveen-garg: Yes please; make consistent as 2px

praveen-garg (dismissed):

mshariq-nerd (approved):

hemantgarg923 (approved):

PR #464, opened 2024-09-25
FOGL-9129: North complex pipeline does not allow to choose a filter

hemantgarg923:


src/app/components/common/node-editor/editor.ts
```diff
@@ -335,25 +335,29 @@ export function getUpdatedFilterPipeline() {
...
     else {
+      // check if all the filters are connected from both sides
       if (!connections.find(c => c.source === nodes[i].id) || !connections.find(c => c.target === nodes[i].id)) {
         console.log("Dangling connection");
         return false;
       }
     }
   }
 
+  // check for self loop
```
praveen-garg, line 360: self loop?

praveen-garg (approved):

src/app/components/common/node-editor/editor.ts
hemantgarg923: This check is to test if there is a connection from input socket to output socket of same node.

src/app/components/common/node-editor/insert-node/index.ts
```diff
@@ -94,8 +94,10 @@ async function removeOldConnection(node, editor) {
       inputConnId = element.id;
     }
   }
-  for (let t of target) {
-    await editor.addConnection(new Connection(connectionEvents, source, t));
+  if(source) {
+    for (let t of target) {
```
MonikaSharma06, line 98: It should be like-

`for (let target of targets) {`

MonikaSharma06 (dismissed):

src/app/components/common/node-editor/insert-node/index.ts
praveen-garg: agree `targets` as plural!

`for (let target of targets) {`  or `for (let t of targets) {`

src/app/components/common/node-editor/editor.ts
```diff
@@ -335,25 +335,29 @@ export function getUpdatedFilterPipeline() {
 
   for (let i = 0; i < nodes.length; i++) {
     if (i == 0) {
+      // check if starting node of pipeline i.e. south/storage is connected
       if (!connections.find(c => c.source === nodes[i].id)) {
         console.log("Dangling connection");
```
praveen-garg, line 340: not in the scope of this, but should we log more details in logs? Dangling connection for or from .. to ...

praveen-garg (dismissed):

mshariq-nerd (approved):

mshariq-nerd (approved):

PR #468, opened 2024-10-16
FOGL-8975: Added tab navigator buttons for filter plugins configuration

MonikaSharma06:


src/app/components/core/configuration-manager/configuration-group/configuration-group.component.html
```diff
@@ -9,8 +9,8 @@
     </ul>
   </nav>
   <div class="configuration_group_wrapper">
-    <nav id="group_navigation" class="groupNav dragscroll mouse-scroll" role="tablist">
-      <div id="groupNavContents" class="nav" [ngClass]="{'groupNav_contents': !sourceName}">
+    <nav [id]="'group_navigation_'+from + '_' + sourceName" class="groupNav dragscroll mouse-scroll" role="tablist">
+      <div [id]="'groupNavContents_'+from + '_' + sourceName" class="nav">
```
praveen-garg, line 13: We have nav before as tag and here as a class?

src/app/components/core/configuration-manager/configuration-group/configuration-group.component.html
```diff
@@ -9,8 +9,8 @@
     </ul>
   </nav>
   <div class="configuration_group_wrapper">
-    <nav id="group_navigation" class="groupNav dragscroll mouse-scroll" role="tablist">
-      <div id="groupNavContents" class="nav" [ngClass]="{'groupNav_contents': !sourceName}">
+    <nav [id]="'group_navigation_'+from + '_' + sourceName" class="groupNav dragscroll mouse-scroll" role="tablist">
```
praveen-garg, line 12: In next line ID convention is camelCase and here snake_case?

src/app/components/core/configuration-manager/configuration-group/configuration-group.component.ts
```diff
@@ -17,6 +17,7 @@ export class ConfigurationGroupComponent implements AfterViewInit {
   @Input() serviceStatus = false;
   @Input() from: string;
   @Input() sourceName: string;
+  @Input() isFilterConfigOpen: boolean;
```
praveen-garg, line 20: You mean expanded? How expanded and collapsed matter.

src/app/components/core/configuration-manager/configuration-group/configuration-group.component.ts
```diff
@@ -147,7 +155,7 @@ export class ConfigurationGroupComponent implements AfterViewInit {
 
   buildGroupOfItems(configItems) {
     configItems?.forEach(config => {
-      if(config.readonly != 'true'){
```
praveen-garg, line 150: What if readonly does not exist?

src/app/components/core/filter/filter-list/filter-list.component.ts
```diff
@@ -78,15 +79,18 @@ export class FilterListComponent {
         const nextActiveContentBody = <HTMLElement>next.getElementsByClassName('card-content')[0];
         nextActiveContentBody.hidden = false;
         next.setAttribute('class', 'accordion card is-active');
+
```
praveen-garg, line 82: Required?

src/app/components/core/filter/filter-list/filter-list.component.ts
```diff
@@ -78,15 +79,18 @@ export class FilterListComponent {
...
       } else {
         last.classList.remove('is-active');
         lastActiveContentBody.hidden = true;
       }
+      this.isConfigOpen = false;
```
praveen-garg, line 87: Should be in else, above?

src/styles.css
```diff
@@ -507,4 +507,9 @@ span.tooltip {
 .content table td {
   padding: 0.25em 0.5em;
 }
+
+.steps-content {
```
praveen-garg, line 511: In wizard no affect?

src/app/components/core/configuration-manager/configuration-group/configuration-group.component.html
MonikaSharma06: No issue in that. It is same as before. I only removed custom class i.e. `groupNav_contents` because property (which was being applied in class `groupNav_contents`) needs to be applied on for all.

src/app/components/core/configuration-manager/configuration-group/configuration-group.component.ts
MonikaSharma06: We need to do some calculation to show tab navigator buttons. In the case of filter, it was not working properly because by default configuration (and tabs) are hidden and get visible when we expand it. So we need to calculate it again when user expand the filter.

src/app/components/core/configuration-manager/configuration-group/configuration-group.component.ts
MonikaSharma06: This line is not removed from here, there is only indentation fix (by editor), please check.

src/styles.css
MonikaSharma06: No. It is required to show tab navigator on wizard.

src/app/components/core/configuration-manager/configuration-group/configuration-group.component.html
MonikaSharma06: ID convention updated

hemantgarg923 (dismissed):

mshariq-nerd (approved):

hemantgarg923 (approved):

PR #470, opened 2024-10-17
FOGL-9108: Fixed integer and float directive to accept valid input

MonikaSharma06:


src/app/directives/float-only.directive.ts
```diff
@@ -1,23 +1,72 @@
...
+    }
+  }
+
+  isVaildFloat(event) {
+    const allowedChars = /[0-9.]/;
+    const inputChar = String.fromCharCode(event.charCode);
+    const isAllowedKeys = ['Backspace', 'Tab', 'ArrowLeft', 'ArrowRight', 'Delete', 'e', 'E', '-', '+'].includes(event.key);
+    const isScientificNotation = event.key.toLowerCase() === 'e';
+    const hasAlreadyScientificNotation = this.el.nativeElement.value.includes('e') || this.el.nativeElement.value.includes('E');
+    if (allowedChars.test(inputChar) || isAllowedKeys) {
```
praveen-garg, line 39: Here we have regex check for  /[0-9.]/; Cannot we have same for int?
or
event.key here?

praveen-garg (commented):
Can we have a directive number-only and have common code there; Override that for int and float?

src/app/directives/float-only.directive.ts
```diff
@@ -1,23 +1,72 @@
...
     if (inputChar === '.' && (event.target as HTMLInputElement).value.includes('.')) {
       event.preventDefault();
     }
+
+    if (event) {
+      return this.isVaildFloat(event);
+    }
+  }
+
+  isVaildFloat(event) {
```
hemantgarg923, line 33: `isValidFloat` typo

src/app/directives/float-only.directive.ts
MonikaSharma06: Yes, we can have same also.

> Can we have a directive number-only and have common code there; Override that for int and float?

Yes, we can try in that way also but to keep these separate seems better.

hemantgarg923 (dismissed):

praveen-garg:
> Can we have a directive number-only and have common code there; Override that for int and float?

@mshariq-nerd thoughts?

hemantgarg923 (approved):

mshariq-nerd:
> > Can we have a directive number-only and have common code there; Override that for int and float?
> 
> @mshariq-nerd thoughts?

Either way is fine. If we will have a number-only directive in that from every configuration item of type int & float user need to pass the type as well to check if float or number.

mshariq-nerd (approved):

PR #474, opened 2024-10-24
FOGL-9212: Show filter delete confirmation message

MonikaSharma06:


hemantgarg923 (approved):

praveen-garg:
[commit/83535af](https://github.com/fledge-iot/fledge-gui/commit/83535af2824d9ae69d00e5657c59341b6be17bb3#diff-3a2b7491de1f0545417194779cc18f8a146a4b7f4df4be3dba93f5439bcee905L227) says
- delete not attached filters silently in control pipeline

Are we reverting that behavior? or We need this to be South & North vs Control Pipeline specific?
Also, at some places we are using alertService while toastService at some?

MonikaSharma06:
> Are we reverting that behavior? or We need this to be South & North vs Control Pipeline specific?

We are not reverting any behaviour. I've reviewed code wise and tested too but I don't see any bad effect of it anywhere. FYI; this method is not being called from Control Pipeline component.

> Also, at some places we are using alertService while toastService at some?

Yes, we need to be consistent on it.

mshariq-nerd (approved):

PR #478, opened 2024-11-18
FOGL-9257: Fixed 'Next' button disabled issue on notification wizard

MonikaSharma06:


src/app/components/core/notifications/add-notification-wizard/add-notification-wizard.component.ts
```diff
@@ -171,6 +172,7 @@ export class AddNotificationWizardComponent implements OnInit, OnDestroy {
                 this.pluginData.modalState = false;
                 this.selectInstalledPlugin();
               }
+              this.pluginRequestDone = true;
```
praveen-garg, line 175: 1s timeout enough?

MonikaSharma06: No need to change timeout, only need to enable "Next" button after getting the response from `plugin` API.

hemantgarg923 (approved):

PR #483, opened 2025-01-27
FOGL-9424: Invalid connection validation check for South and North

mshariq-nerd:


src/app/components/common/node-editor/connector.ts
```diff
@@ -14,8 +17,26 @@ export class Connector<S extends ClassicScheme, K extends any[]> extends Bidirec
...
         const toNode = context.editor.getNode(socket.nodeId);
+        const isNorthNode = context.editor.getNodes().find(node => node.label == 'North');
+
+        // Invalid connection check
+        if (
+          (fromNode instanceof Storage && toNode instanceof South) ||
+          (fromNode instanceof North && toNode instanceof Storage) ||
+          (fromNode instanceof Filter && toNode instanceof South) ||
+          (fromNode instanceof North && toNode instanceof Filter) ||
+          (isNorthNode && fromNode instanceof Filter && toNode instanceof Storage)
```
praveen-garg, line 35: I would rather have a validConnections array of objects with src, dst and if does not have any return with console log

hemantgarg923 (approved):

MonikaSharma06 (approved):

PR #484, opened 2025-01-27
FOGL-9300: Import CSV/ JSON file support for list & kvlist

hemantgarg923:


mshariq-nerd (changes requested):

src/app/components/common/file-export-modal/file-export-modal.component.ts
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  toggleDropDown(id: string) {
+    const dropdown = document.getElementById(id);
+    if (dropdown) {
+      dropdown.classList.toggle('is-active');
+    }
+  }
+
+  hideDropDown(id: string) {
```
praveen-garg, line 56: Given we have `toggleDropDown`, either we should not have `hideDropDown` or instead of toggleDropDown use `showDropdown`

hemantgarg923: Currently both methods are required as they serve different purpose, maybe we can improve it in a patch PR.

praveen-garg: OK

mshariq-nerd (approved):

PR #485, opened 2025-02-04
FOGL-7733: Reset password option hidden for certificate based user

hemantgarg923:


praveen-garg:
Only problem here is, what if reset password is required before switching to ANY/PASSWD option.

PR #489, opened 2025-02-06
FOGL-9482: Handling of listName for list type configuration item

hemantgarg923:
Check with [PR](https://github.com/fledge-iot/fledge/pull/1549)


praveen-garg:
list and kvlist both are tested with it?

We may want to check S2 OPC UA like JSON to list/kvlist migration pull requests?

hemantgarg923:
> GUI client should handle the case where a configuration item contains a listName key-value pair in case of list type configuration only. 

As per Jira description, listName handling is done for **list** type only.

praveen-garg:
> > GUI client should handle the case where a configuration item contains a listName key-value pair in case of list type configuration only. 
> 
> 
> 
> As per Jira description, listName handling is done for **list** type only.

Does this mean, We don't need for KVList? I doubt that.

ashish-jabble (approved):

MonikaSharma06 (approved):

mshariq-nerd (approved):

PR #495, opened 2025-02-20
FOGL-9512: Changes to make flow editor default view

MonikaSharma06:
Along with JIRA scope, handled below points also-

- [x] Fix issue- Save Cancel is sticking to bottom on notification instances details page
- [x] Removed border under the selected tab

### DIFF
--- a/src/app/components/core/settings/settings.component.ts
+++ b/src/app/components/core/settings/settings.component.ts
@@ -49,7 +49,9 @@ export class SettingsComponent implements OnInit {
     this.protocol = this.storageService.getProtocol() != null ? this.storageService.getProtocol() : location.protocol.replace(':', '').trim();
     this.host = this.storageService.getHost() != null ? this.storageService.getHost() : location.hostname;
     this.servicePort = this.storageService.getPort() != null ? this.storageService.getPort() : 8081;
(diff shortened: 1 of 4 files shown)


src/app/components/core/settings/settings.component.ts
```diff
@@ -49,6 +49,9 @@ export class SettingsComponent implements OnInit {
     this.protocol = this.storageService.getProtocol() != null ? this.storageService.getProtocol() : location.protocol.replace(':', '').trim();
     this.host = this.storageService.getHost() != null ? this.storageService.getHost() : location.hostname;
     this.servicePort = this.storageService.getPort() != null ? this.storageService.getPort() : 8081;
+    if (localStorage.getItem('FLOW_EDITOR') == null) {
```
hemantgarg923, line 52: Better to handle this in `flowEditorService` as this will not trigger if the user does not come to settings component.

MonikaSharma06: Whenever user open the application (very first time) he first re-directs to settings page, So I used it here itself.
But now made changes as suggested as it is a better way to handle this in `flowEditorService`.

hemantgarg923 (approved):

mshariq-nerd (approved):

PR #500, opened 2025-03-05
FOGL-9573: Save button, reload button and blank page issues fixed

hemantgarg923:


src/app/components/common/node-editor/node-editor.component.ts
```diff
@@ -1123,11 +1127,7 @@ export class NodeEditorComponent implements OnInit {
     }
     this.filterPipeline = [];
     this.updatedFilterPipeline = [];
-    if (this.task || this.service) {
-      this.router.navigate(['/flow/editor', this.from, this.source, 'details']);
-      return;
```
praveen-garg, line 1128: What was this check for?

hemantgarg923: This check was incorrect as we are showing toolbar only on detail page, so reload button should never redirect to list page

mshariq-nerd (approved):

firozalikhan (approved):

MonikaSharma06 (approved):

PR #505, opened 2025-03-20
FOGL-9626: Truncates long names in the node to fix the issue

MonikaSharma06:


src/app/pipes/center-truncate.pipe.ts
```diff
@@ -0,0 +1,14 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+    name: 'centerTruncate'
+})
+export class CenterTruncatePipe implements PipeTransform {
+    transform(value: string, maxLength: number): string {
```
praveen-garg, line 7:
```
transform(value: string, maxLength: number): string {
    if (!value || value.length <= maxLength) return value;
    const halfLength = Math.floor((maxLength - 3) / 2); // Subtract 3 for "..."
    return value.slice(0, halfLength) + '...' + value.slice(-halfLength);
}
```

src/app/pipes/center-truncate.pipe.ts
```diff
@@ -0,0 +1,14 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+    name: 'centerTruncate'
```
praveen-garg, line 4: truncateMiddleOfString or truncateMiddleOfText

To clearly convey the action of truncating the middle section of the string/text with the preservation of both ends

src/app/pipes/center-truncate.pipe.ts
praveen-garg: Even better to define `function transform(value: string, maxLength: number = 16): string` to avoid passing `:[16]` at two places unless needed.

praveen-garg (approved):

hemantgarg923 (approved):

mshariq-nerd (approved):

PR #509, opened 2025-04-21
FOGL-9711: Save button enable issue fixed on configuration tab

hemantgarg923:


src/app/components/core/configuration-manager/configuration-manager.component.html
```diff
@@ -50,8 +50,8 @@
                 <div class="column">
                   <div class="columns">
                     <div class="column has-text-right">
-                      <button [disabled]="!validConfigForm" class="button is-small is-outlined is-link"
-                        (click)="save(category.name, category.description)"
+                      <button [disabled]="checkFormState() || !validConfigForm"
```
shariq1501: What is the use of validConfigForm? I am not seeing any change in the validConfigForm in component ts file. If this is not required remove or we can use only validConfigForm variable to handle the checkFormState?

mshariq-nerd (changes requested):

src/app/components/core/configuration-manager/configuration-manager.component.html
hemantgarg923: `validConfigForm` variable is being used to check the validity of form, it is being assigned in HTML itself ( check [here](https://github.com/fledge-iot/fledge-gui/blob/99da930ff2dc2bb392ebd46d4e4d113fc143a790/src/app/components/core/configuration-manager/configuration-manager.component.html#L45) ). I have used a separate variable `isFormValueChanged` to hold info about changes in the form and removed function `checkFormState` for better memory management.

mshariq-nerd (approved):

src/app/components/core/configuration-manager/configuration-manager.component.ts
```diff
@@ -168,6 +169,7 @@ export class ConfigurationManagerComponent implements OnInit {
   getChangedConfig(changedConfiguration: any, category: any) {
     const cat = this.categoryDataCopy.find(cat => cat.key === category.key);
     this.changedConfig = this.configurationControlService.getChangedConfiguration(changedConfiguration, cat);
+    this.isFormValueChanged = !isEmpty(this.changedConfig);
```
praveen-garg, line 172: Cannot we use this condition itself; instead of introducing another variable?

Or there are some performance/standard practices concern?

firozalikhan (approved):

MonikaSharma06 (approved):

src/app/components/core/configuration-manager/configuration-manager.component.ts
hemantgarg923: We can use this condition directly in HTML file but in that case `isEmpty` function will be called frequently (on every change detection) which is slightly less memory optimised.

PR #511, opened 2025-04-21
FOGL-9567: Debugger state feedback

mshariq-nerd:


src/app/components/core/debugger/debugger.component.html
```diff
@@ -0,0 +1,238 @@
...
+
+      <hr>
+
+      <form [formGroup]="formGroup">
+        <div class="columns is-mobile mb-3">
+          <div class="column">
+            <label class="label is-small">
+              Buffer <i class="bi bi-xs bi-info-circle info-icon" aria-hidden="true"
+                title="Set the number of readings to hold in each buffer, passing an integer argument"></i></label>
+            <input class="input is-small" type="text" formControlName="buffer" appIntegerOnly>
```
praveen-garg: Cannot we use type number with min=1 step=1?

src/app/components/core/debugger/debugger.component.html
```diff
@@ -0,0 +1,238 @@
...
+            <p class="is-size-7 has-text-grey mt-1" *ngIf="bufferStatus === 'saving'">Saving...</p>
+            <p class="is-size-7 has-text-success mt-1" *ngIf="bufferStatus === 'saved'">Saved!</p>
+          </div>
+          <div class="column">
+            <label class="label is-small">
+              Steps Reading
+              <i class="bi bi-xs bi-info-circle info-icon" aria-hidden="true"
+                title="Allow readings to flow into the pipeline. Passed an optional number of readings to ingest; default to 1 if omitted">
```
praveen-garg: Pass

src/app/components/core/debugger/debugger.component.html
```diff
@@ -0,0 +1,238 @@
...
+              </i>
+            </label>
+
+            <div class="field has-addons">
+              <p class="control is-expanded">
+                <input #step class="input is-small" type="text" formControlName="step" appIntegerOnly>
```
praveen-garg: As above; do we need formControlName when used #step already

praveen-garg: Can steps be larger than buffer size?

src/app/components/common/node-editor/custom-connection/custom-connection.component.html
```diff
@@ -2,10 +2,8 @@
   <path [attr.d]="path" class="hover-path"></path>
   <path [attr.d]="path" #svgpath [class]="{ selected: data.selected }" />
 </svg>
-<!-- Show label on the connection line -->
-<!-- <span (pointerdown)="$event.stopPropagation()" class="connection-status label has-text-success" [style]="{
-    left: point.x + 'px',
-    top: point.y + 'px',
-  }">
-  <i class="bi bi-vinyl"></i>
+<!-- Show ban icon on the connection wire, if egress/ingress disabled -->
```
praveen-garg: Remove this and instead show a most recent reading icon with popover having reading at each stage?

src/app/components/common/node-editor/custom-node/custom-node.component.css
```diff
@@ -118,11 +118,6 @@
   padding: 0px 4px 4px 4px;
 }
 
-.reading-details {
```
praveen-garg, line 121: What was this for?

src/app/components/common/node-editor/custom-node/custom-node.component.css
```diff
@@ -234,4 +229,78 @@
...
+  left: 48%;
+  transform: translate(-50%, -50%);
+  color: gray;
+  font-size: xx-large !important;
+}
+
+.ingress-isolated {
+  position: absolute;
+  top: 45%;
+  left: 44%;
```
praveen-garg: I think we can have 


.isolated .egress 
And move common to isolated?

src/app/components/common/node-editor/custom-node/custom-node.component.css
```diff
@@ -234,4 +229,78 @@
...
+
+.egress-isolated {
+  position: absolute;
+  top: 45%;
+  left: 48%;
+  transform: translate(-50%, -50%);
+  color: gray;
+  font-size: xx-large !important;
+}
+
```
praveen-garg, line 251: What is ingress isolation?

src/app/components/common/node-editor/node-editor.component.ts
```diff
@@ -63,10 +63,14 @@ export class NodeEditorComponent implements OnInit {
   private nodeClickSubscription: Subscription;
   private nodeDropdownClickSubscription: Subscription;
   private pipelineSubscription: Subscription;
+  private debuggerStateSubscription: Subscription;
 
   showPluginConfiguration: boolean = false;
   showFilterConfiguration: boolean = false;
   showLogs: boolean = false;
+  debugger: any = {};
+  debuggerPage = false;
```
praveen-garg, line 75: Page or Pane/QuickView/Overlay

MarkRiddoch (changes requested):
This has gone beyond the scope of this Jira and has implemented some of the things for which a mockup was requested in https://scaledb.atlassian.net/browse/FOGL-9630. That other Jira has content that would have been relevant in doing the extra work that has been done here and references a requirements spec and the spec of the API that would have been useful in doing this work [Fledge/FogLAMP User Interface  For The Data Pipeline Debugger](https://docs.google.com/document/d/1cB7U8rSDDwE3NxqDvKjJ0r_Qsp8_F6lpAPxQ06egy2s/edit?tab=t.0#heading=h.qu7rrjl3pr1q).

Having seen the demo there are a couple of things that are not really right.

1. The handling of the step command. Step is an operation that takes an argument of the number of readings to ingest. Having a simple entry box for the step and no action button or menu item does not really work.
2. Displaying the readings in the side panel is not very helpful to the user. The Jira referenced above describes showing the list of readings attach t the nodes that the readings apply to. This is far more useful to the user.

The use of the side panel is not really ideal, it takes a lot of space that prevents the user seeing the pipeline that are debugging. Moving the data display to the nodes rather than in the side panel would mean there is very little to actually show in the panel. It would probably be better at that point to get rid of it completely and find a different way to access what is in that panel.

mshariq-nerd:
@MarkRiddoch, 
>The handling of the step command. Step is an operation that takes an argument of the number of readings to ingest. Having a simple entry box for the step and no action button or menu item does not really work.
  
You may have missed the action was on input; however an explicit action button is added as steps value change is not always expected.
Screenshot 2025-04-22 at 19 55 13

>Displaying the readings in the side panel is not very helpful to the user. The Jira referenced above describes showing the list of readings attach t the nodes that the readings apply to. This is far more useful to the user.

This is what I’m currently working but parsing the buffered data JSON is hard and not following any standard. Here each node wire will show reading as traveling from South to Storage and Storage to North via filters.
Did you read the JSON I had shared in the JIRA 
```
{
            "name": "threshold",
            "readings": [...]
        },
        {
            "name": "delta",
            "readings": []
        },
        {
            "name": "Writer",
            "readings": []
        }
```
a) I don't see a branch for this and this is not inside an array.
b) In data array there are 2 branch objects and followed by array however you can not see their association, how branch object is associated with Filters and Writer object in the following array. 
c) One of the branch is sinusoid -> exp -> storage, why do we need 3 reading outputs as branch, expression & Writer. The output of expression is not being written to storage? Here expected are 2 readings one from sinusoid plugin and another processed by expression which will be written to storage. Same pattern is with other branches.

praveen-garg (commented):
Please remove the commented code which is out of the scope of this Jira.

mshariq-nerd:
@MarkRiddoch,  @brycekroencke  This PR is updated as per the JIRA scope now. Please have a look.

MarkRiddoch:
This gives the basic feedback, so is what we need at least to start with. Bryce will mockup the actual interaction at some point. One minor issue, the tooltip shows underneath other south services, possibly also true for north.
Screenshot 2025-07-03 at 11 21 59
Not clear if auto refresh is happening in all cases. It seems to work when in the page showing all the south services. However in the detailed view, doing a suspendIngest and a resumeIngest it seems to pick it up dynamically. But isolate and store don't. We should be consistent with these.

MarkRiddoch (approved):
Testing this again the previous issues are resolved.

PR #516, opened 2025-05-09
[FOGL-9677] partB: Fix for loading issue on ML model page

MonikaSharma06:


src/app/components/core/developer/additional-services/additional-services-utils.service.ts
```diff
@@ -83,7 +83,7 @@ export class AdditionalServicesUtils {
         (data: any) => {
           let servicesRegistry = data.services.filter((el) => expectedExternalServiceType.includes(el.type))
           if (from !== 'additional-services') {
-            servicesRegistry = data.services.filter((s => (s.type.toLowerCase() === from)));
+            servicesRegistry = data.services.filter((s => (s.type.toLowerCase() === (from === 'bucket' ? 'bucketstorage' : from))));
```
praveen-garg: bucket is not additional-service? the condition above says `if (from !== 'additional-services') `

praveen-garg: We better have clarity in statements as

const origin = from === 'bucket' ? 'bucketstorage' : from;
const servicesRegistry = data.services.filter(service => service.type.toLowerCase() === origin);

MonikaSharma06: Bucket is an additional service but here we are checking from where this function is being called i.e. from `additional-services` page or any other service page like- `notification`, `bucket` etc.

praveen-garg (approved):

mshariq-nerd (approved):

hemantgarg923 (approved):

PR #521, opened 2025-05-23
FOGL-9875 Add a search option in the graph select dropdown in the dashboard

MarkRiddoch:
@mshariq-nerd I would be interested in your view on the code.  Not mine, but the functionality seems useful.

## What do these changes do?

Add a search facility to the drop down selection of graphs in the dashboard

## Are there changes in behavior for the user?

Makes finding the gaps you want easier when the list of graphs is very long
## Checklist

- [ ] Code is well written w.r.t. best practices and proper comments
- [ ] Unit tests for the changes exist
- [ ] e2e tests for the changes exist
- [ ] Documentation reflects the changes including changelog
- [ ] Build works with nginx-light deployment
- [ ] Build works with Docker, if any change on docker related config files

### DIFF
--- a/src/app/components/core/dashboard/dashboard.component.html
+++ b/src/app/components/core/dashboard/dashboard.component.html
@@ -4,7 +4,7 @@
       <div class="field-body is-pulled-right">
         <div id="time-dropdown" class="dropdown is-left">
           <div class="dropdown-trigger">
-            <button class="button is-rounded" aria-haspopup="true" aria-controls="dropdown-menu"
+            <button class="button is-rounded" aria-haspopup="true" aria-controls="time-dropdown-menu"
               (click)="toggleDropDown('time-dropdown')">
               <span>{{optedTime}} minutes</span>
               <span class="icon is-small">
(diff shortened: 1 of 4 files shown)


mshariq-nerd:
@MarkRiddoch,  This looks good overall, but there's one concern — we already have a filter pipe in the codebase. To avoid a naming conflict, we should rename this new pipe to something more specific (e.g., searchFilterPipe).

Also, instead of using the pipe directly, we should declare and export it through the shared PipesModule to ensure consistent reuse across the application. 

With these changes in place, we can go ahead and merge this. It would also be a good idea to made these changes with a JIRA issue for better visibility and tracking.

MarkRiddoch:
> @MarkRiddoch, This looks good overall, but there's one concern — we already have a filter pipe in the codebase. To avoid a naming conflict, we should rename this new pipe to something more specific (e.g., searchFilterPipe).
> 
> Also, instead of using the pipe directly, we should declare and export it through the shared PipesModule to ensure consistent reuse across the application.
> 
> With these changes in place, we can go ahead and merge this. It would also be a good idea to made these changes with a JIRA issue for better visibility and tracking.

Thanks Shariq. This was an experiment using one of the AI code generators to see how good, or otherwise, a job it did. I'll feed it with your comments and crate a real Jira for it. I think it is a useful addition to the interface.

PR #522, opened 2025-05-23
Increased max threshold limit of listSize for new item creation

MonikaSharma06:


praveen-garg (approved):

src/app/components/core/configuration-manager/kv-list-type-configuration/kv-list-type-configuration.component.ts
```diff
@@ -93,7 +93,7 @@ export class KvListTypeConfigurationComponent implements OnInit {
 
   addListItem(isPrepend) {
     const controlsLength = this.kvListItems.length;
-    const listSize = this.configuration?.listSize > 0 ? +this.configuration.listSize : 999; // max threshold limit for new item creation
+    const listSize = this.configuration?.listSize > 0 ? +this.configuration.listSize : 9999; // max threshold limit for new item creation
```
hemantgarg923, line 96: Better to use even higher number for this as there could be 10000 items in the list. We should use 2147483647 ( 32 bit Max integer ) for safe side.

praveen-garg: 999 was too low
9999 is fine.

With other Jira in place this hard coded limit will go away. 

Ideally 99999 is a good choice for these.

hemantgarg923 (approved):

PR #525, opened 2025-06-09
FOGL-9527: Removed confirmation prompt when delete a connection from Delete key

MonikaSharma06:


mshariq-nerd (dismissed):

src/app/components/common/node-editor/node-editor.component.ts
```diff
@@ -1309,19 +1308,20 @@ export class NodeEditorComponent implements OnInit {
...
+  async onDeleteAction() {
     const connectionToDelete = this.nodesToDelete.find(node => (node.label === 'Connection'));
     if (connectionToDelete) {
       await deleteConnection(connectionToDelete.id);
       this.nodesToDelete = [];
       // check if filter pipeline is updated and emit the updated pipeline
       const pipeline = getUpdatedFilterPipeline();
       this.flowEditorService.emitPipelineUpdate(pipeline);
+    } else {
+      this.openModal('from-toolbar-dialog');
```
praveen-garg: better to use named-parameters-in-typescript. then you just can pass from: "toolbar-dialog"? BTW what is toolbar-dialog?

src/app/components/common/node-editor/node-editor.component.ts
```diff
@@ -1309,19 +1308,20 @@ export class NodeEditorComponent implements OnInit {
     }
   }
 
-  onDeleteAction() {
-    this.openModal('from-toolbar-dialog');
-  }
-
-  async callDeleteAction() {
+  async onDeleteAction() {
```
praveen-garg: should be onDelete

src/app/components/common/node-editor/node-editor.component.ts
```diff
@@ -1309,19 +1308,20 @@ export class NodeEditorComponent implements OnInit {
...
     const connectionToDelete = this.nodesToDelete.find(node => (node.label === 'Connection'));
     if (connectionToDelete) {
       await deleteConnection(connectionToDelete.id);
       this.nodesToDelete = [];
       // check if filter pipeline is updated and emit the updated pipeline
       const pipeline = getUpdatedFilterPipeline();
       this.flowEditorService.emitPipelineUpdate(pipeline);
+    } else {
```
praveen-garg: add inline comment? what is ! connectionToDelete thing?

src/app/components/common/node-editor/node-editor.component.ts
```diff
@@ -1309,19 +1308,20 @@ export class NodeEditorComponent implements OnInit {
...
+      this.openModal('from-toolbar-dialog');
     }
+  }
+
+  async callDeleteAction() {
```
praveen-garg: Why not just async delete or async deleteNodes? I'm not sure on naming convention here.

src/app/components/common/node-editor/node-editor.component.ts
MonikaSharma06: In this component, "toolbar-dialog" is the confirmation dialog which open when clicking on the delete button from the toolbar.

src/app/components/common/node-editor/node-editor.component.ts
MonikaSharma06: It is the connection which is selected and need to be deleted, BTW; changed its name to "selectedConnection"

src/app/components/common/node-editor/node-editor.component.ts
MonikaSharma06: changed the name

hemantgarg923 (approved):

praveen-garg (commented):
I don't see any change in closeModal for 'from-toolbar-dialog' vs 'toolbar-dialog'>

mshariq-nerd (approved):

MonikaSharma06:
> I don't see any change in closeModal for 'from-toolbar-dialog' vs 'toolbar-dialog'>

'closeModal' handled in both (`ts` and `html`) files for 'toolbar-dialog'

PR #543, opened 2025-07-22
Show Applications text in the middle of hr; and remove extra hr if fi…

praveen-garg:
…lters length > 0; Also show help text when no filters configured

Screenshot 2025-07-23 at 3 28 46 AM

Vs

Screenshot 2025-07-23 at 3 28 53 AM

Also when filter is there, no extra line following the box line

Screenshot 2025-07-23 at 3 29 06 AM

### DIFF
--- a/src/app/components/core/north/north-task-modal/north-task-modal.component.html
+++ b/src/app/components/core/north/north-task-modal/north-task-modal.component.html
@@ -83,6 +83,7 @@
           </div>
           <div class="hr-container">
             <hr class="hr-text">
+            <!-- this should vainsh on ends-->
             <div class="hr-text-content">
               <span class="label is-size-6 add-application" *ngIf="rolesService.hasEditPermissions(); else staticLabel">
                 <a (click)="openAddFilterModal(true)">Applications
--- a/src/app/components/core/north/north-task-modal/north-task-modal.component.html
+++ b/src/app/components/core/north/north-task-modal/north-task-modal.component.html
@@ -92,14 +92,14 @@
                 </a>
               </span>
               <ng-template #staticLabel>
-                <span class="label is-size-6">Applications</span>
+                <span class="label is-size-6 add-application">Applications</span>
               </ng-template>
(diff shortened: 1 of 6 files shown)


src/app/components/core/north/north-task-modal/north-task-modal.component.html
```diff
@@ -98,7 +98,10 @@
           </div>
           <app-filter-list #filtersListComponent [service]="task?.name" [filterPipeline]="filterPipeline"
             [from]="'filter'" (formStatus)="filterFormStatus($event)"></app-filter-list>
-          <hr *ngIf="rolesService.hasEditPermissions()" class="hr-text">
```
praveen-garg, line 101: Why this condition was for north and not south?!

src/app/components/core/south/south-service-modal/south-service-modal.component.html
```diff
@@ -46,13 +46,18 @@
                 </a>
               </span>
               <ng-template #staticLabel>
-                <span class="label is-size-6">Applications</span>
+                <span class="label is-size-6 add-application">Applications</span>
```
praveen-garg: Why did not we have `add-application` class here; different colour for different views? I intentionally left for north and may be we need some typography class instead of this extra (badly named) class and css with opacity?

src/app/components/core/north/north-task-modal/north-task-modal.component.html
MonikaSharma06: Maybe we missed on south.

src/app/components/core/south/south-service-modal/south-service-modal.component.html
MonikaSharma06: > we need some typography class instead of this extra (badly named) class and css with opacity

Right

MonikaSharma06 (approved):

hemantgarg923 (approved):

PR #544, opened 2025-07-31
FOGL-10080: Integrate bulma popover extension

hemantgarg923:


praveen-garg (commented):
How long it requires to show popover? My concern is that it's neither too quick nor too slow.
When something is loading do we show loading.... muted text as popover content?
We should move common CSS to popover css file?

src/app/components/core/asset-readings/assets/assets.component.css
```diff
@@ -40,3 +40,54 @@
...
+  background: #a8a8a8;
+}
+
+.popover.is-popover-right .popover-content {
+  transform: translate(0.2rem, 95%) !important;
+}
+
+.popover .popover-content::before {
+  border-width: 0rem;
+}
```
MonikaSharma06, line 46: Please check if duplicate css can be removed.

praveen-garg (dismissed):

hemantgarg923:
Closing with reference of [https://github.com/fledge-iot/fledge-gui/pull/552](https://github.com/fledge-iot/fledge-gui/pull/552)

PR #562, opened 2025-09-09
Image and object type reading shown in popover

hemantgarg923:


src/app/components/core/asset-readings/assets/assets.component.css
```diff
@@ -68,3 +68,65 @@ table.scroll tr {
...
+
+.popover-scrollable-content .table td {
+  border: none;
+  padding: 0.25rem 0.5rem;
+  vertical-align: middle;
+}
+
+.popover-scrollable-content .table tr:nth-child(even) {
+  background-color: rgba(0, 0, 0, 0.02);
+}
```
shariq1501: Popover related css should be in the popover component?

src/app/components/core/asset-readings/assets/assets.component.ts
```diff
@@ -132,21 +132,140 @@ export class AssetsComponent implements OnInit, OnDestroy {
...
+      imgData.data[i + 2] = num_scaled_8;
+      imgData.data[i + 3] = 255;
+      x += 2;
+    }
+
+    ctx.putImageData(imgData, 0, 0);
+    return canvas.toDataURL("image/png");
+  }
+
+  private process24bitBitmap(buffer: ArrayBuffer, options: { width: number, height: number }): string {
```
shariq1501: This is a repeated code. We already have processImageReading, process24bitBitmap and process8bitBitmap function in readings-graph.component. Better to move the repeted code in a service and use it on both places.

mshariq-nerd (changes requested):

gnandan (approved):
Reviewed by testing

mshariq-nerd (approved):

PR #566, opened 2025-09-30
Complex filter detection tabular mode

praveen-garg:
Screenshot 2025-10-01 at 2 44 52 AM

### DIFF
--- a/src/app/components/core/north/north-task-modal/north-task-modal.component.ts
+++ b/src/app/components/core/north/north-task-modal/north-task-modal.component.ts
@@ -22,12 +22,7 @@ import { Subject, forkJoin, of } from 'rxjs';
 import { catchError, map, takeUntil } from 'rxjs/operators';
 import { NorthTask } from '../north-task';
 import { FilterListComponent } from '../../filter/filter-list/filter-list.component';
-
-export enum FilterPipelineType {
-  Empty = 'empty',
-  Simple = 'simple',
-  Complex = 'complex'
-}
+import { FilterPipelineType } from '../../../../services/filter.service';
 
 @Component({
   selector: 'app-north-task-modal',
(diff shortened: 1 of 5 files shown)


src/app/components/core/south/south-service-modal/south-service-modal.component.ts
```diff
@@ -186,6 +190,13 @@ export class SouthServiceModalComponent implements OnInit {
     modalWindow.classList.remove('is-active');
   }
 
+  /**
+   * Get the filter pipeline type based on current pipeline structure
+   */
+  public get filterPipelineType(): FilterPipelineType {
```
hemantgarg923: Pipeline type check is needed only one time, it will not get changed dynamically. Better to create a normal variable instead of a getter property and assign value to that in ngOnInit lifecycle hook to slightly improve the performance.

src/app/components/core/north/north-task-modal/north-task-modal.component.ts
```diff
@@ -268,6 +272,13 @@ export class NorthTaskModalComponent implements OnInit, OnChanges {
     this.repeatTime = event.target.value;
   }
 
+  /**
+   * Get the filter pipeline type based on current pipeline structure
+   */
+  public get filterPipelineType(): FilterPipelineType {
```
hemantgarg923: Pipeline type check is needed only one time, it will not get changed dynamically. Better to create a normal variable instead of a getter property and assign value to that in ngOnInit lifecycle hook to slightly improve the performance.

praveen-garg:
Agree, However

1) What is the performance impact?
2) My thought to see changes which are being made via REST API (cURL or deployment  by centralised management system)

hemantgarg923:
> Agree, However
> 
> 1. What is the performance impact?
> 2. My thought to see changes which are being made via REST API (cURL or deployment  by centralised management system)

1. We generally do not use getter functions approach although it will not cause much performance issue here because we are calling only a utility function inside that.
2. Even after using the getter function we will not be able to see changes made by cURL as we are passing the same pipeline again & again in the function i.e. we are not making an API call to get the updated pipeline from backend.

hemantgarg923 (approved):

mshariq-nerd (approved):

